PR #33, opened 2026-03-11
feat: support Exoscale infrastructure

Gregory-Bull-Exasol:
replaced by https://github.com/exasol/exasol-personal/pull/43

### DIFF
--- a/assets/infrastructure/exoscale/README.md
+++ b/assets/infrastructure/exoscale/README.md
@@ -1,0 +1,125 @@
+# Exoscale Infrastructure as Code Architecture
+
+## Overview
+This document describes the Infrastructure as Code (IaC) implementation for Exasol Personal on Exoscale. It supports both single-node and multi-node (cluster) deployments with a simple, opinionated setup for networking, storage, and installation.
+
+## Prerequisites and Exoscale Provider
+- Exoscale API credentials are taken from environment variables: `EXOSCALE_API_KEY` and `EXOSCALE_API_SECRET`.
+- Zone selection is required via `TF_VAR_zone` (e.g., `ch-gva-2`, `de-fra-1`, `at-vie-1`). Unlike AWS availability zones, Exoscale zones are analogous to AWS regions.
+- For SOS (S3-compatible object storage) access, the same credentials can be mapped to `AWS_ACCESS_KEY_ID` and `AWS_SECRET_ACCESS_KEY` for the `aws.sos` provider.
+- The provider configuration in `providers.tf` uses these environment variables and does not define credentials inline.
+
+## Infrastructure Components
+
+### Compute
+- Exoscale compute instances named after their node IDs (e.g., `n11`, `n12`, ...).
+- Default instance type `standard.extra-large` (4 vCPU / 32 GB RAM; configurable via `var.instance_type`).
+- Ubuntu template: uses `Linux Ubuntu 22.04 LTS 64-bit` from Exoscale's public template catalog.
+- Cluster support: one instance per node; controlled by `var.cluster_size` (default: 1).
+- Instances get public IPv4 addresses automatically; no Elastic IP management required.
+
+### Storage
+- Separate volumes for OS/root and database data.
+- OS disk size configurable via `var.os_volume_size` (minimum 10 GB).
+- Data volumes are separate block storage volumes, encrypted by default, attached via `block_storage_volume_ids`.
+- Data volume size configurable via `var.data_volume_size` (in GB).
+- A remote archive volume on Exoscale SOS (S3-compatible object storage) is created and registered automatically if `var.s3_archive_enabled` is true (default).
+  - A per-deployment SOS bucket is created using a globally unique name.
+  - Access to the bucket is granted via an IAM role and API key pair delivered to instances via cloud-init (no instance profiles; explicit credentials).
+
+### Networking
+- **Simplified Model**: Exoscale instances receive public IP addresses automatically. No VPC, Internet Gateway, or route tables are needed.
+- A managed private network is created for inter-node communication with static DHCP leases.
+- Private IPs are assigned from the `172.30.1.0/24` range.
+- Security groups control inbound/outbound traffic. Exoscale security groups are stateful and use separate rule resources.
+- **No VPC Endpoints**: SOS access goes directly over HTTPS to `sos-<zone>.exo.io`; no private network isolation for object storage.
+
+## Access and Security
+The following ports are exposed via security group rules (configurable per port):
+
+1. 22 — SSH access (public)
+2. 2581 — Default bucketfs (public)
+3. 8443 — Admin UI HTTPS (public)
+4. 8563 — Default database port (public)
+5. 20002 — Exasol container SSH (public)
+6. 20003 — Exasol confd API (public)
+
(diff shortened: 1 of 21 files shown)


assets/infrastructure/exoscale/README.md
nh-exa, line 1: This documentation should not mention AWS or differences to it (except maybe in single section).
I know how this happens, which is why I always tell an LLM to write new documentation in the style of an existing document one but not to change one.

assets/installation/ubuntu/files/opt/exasol_launcher/scripts/prepareExasol.sh
```diff
@@ -33,9 +33,14 @@ chmod +x c4
 log_substep_info "Setting up disk permissions"
 
 data_volume_id="$(node_jq -er '.hostDatadisk')"
+# Truncate to first 20 characters for Exoscale (virtio truncates UUIDs to 20 chars)
+# AWS EBS IDs (vol-xxxxx) are shorter, so this doesn't affect them
+data_volume_id_short="${data_volume_id:0:20}"
 cat <<EOF | tee /etc/udev/rules.d/90-exasol.rules
-SUBSYSTEM=="block", ENV{ID_SERIAL_SHORT}=="${data_volume_id/-/}", OWNER="ubuntu", MODE="0660"
-SUBSYSTEM=="block", ENV{ID_SERIAL_SHORT}=="${data_volume_id/-/}", SYMLINK+="exasol_data_01", MODE="0660"
+# AWS EBS volumes: match on ID_SERIAL_SHORT with hyphen stripped (vol-xxxxx -> volxxxxx)
```
nh-exa: I'd like to avoid having to extend this for every cloud provider.
We could either:
- Find a universal matching rule (might not be possible, though  @moben had an ideas earlier)
- Have in the infrastructure preset write or inject the rule itself via the hook script mechanism (definitively doable)

Also, remember that I said that we are at the 16 kB limit for user-data on AWS?  Adding this much code and comments here will likely break that again.

nh-exa (changes requested):

assets/installation/ubuntu/files/opt/exasol_launcher/scripts/prepareExasol.sh
wizenink: Open for discussion, but even if there is no universal matching rule, we should probably be doing this on `cloudinit.tf`, compute values as tf locals and then write them to env, so the script can get them injected and be more agnostic

nh-exa: Yeah, something like that could work. @wizenink @loaj84 since you also already know whats being done for Azure, maybe you can come up with a proposal that is flexible and extensible so it would work for both Exoscale and Azure as well as future cloud providers. And maybe even consider other Linux distros. We already know we'll want to support RHEL eventually. Would be a shame if build something now we have to re-architecture again for RHEL.

nh-exa: It's probably best to do that in a separate PR and then rebase Exoscale and Azure on that (either before or once it is merged to main).

assets/infrastructure/exoscale/files/opt/exasol_launcher/scripts/exoscale_setup_data_disk.sh
```diff
@@ -0,0 +1,25 @@
...
+source "${SCRIPT_DIR}/config.sh"
+
+log_substep_info "Setting up Exoscale block storage data disk permissions"
+
+data_volume_id="$(node_jq -er '.hostDatadisk')"
+# Truncate to first 20 characters (virtio truncates UUIDs in serial to 20 chars)
+data_volume_id_short="${data_volume_id:0:20}"
+cat <<EOF > /etc/udev/rules.d/90-exasol.rules
+# Exoscale block storage: match on ID_SERIAL (virtio truncates UUIDs to 20 chars)
+SUBSYSTEM=="block", ENV{ID_SERIAL}=="${data_volume_id_short}", OWNER="ubuntu", MODE="0660", SYMLINK+="exasol_data_01"
```
nh-exa, line 20: I didn't think notice that before, but this being part of the infrastructure preset and also ties this to the "ubuntu" user and will fall on our feet at some point too.

As you can see here (Azure PR): https://github.com/exasol/exasol-personal/pull/34/changes#diff-846f5e83d155bdad5f4fd82fd2cb3d8a85534176b508669680f74ca21dde08cc
... which i'm also having my doubts about if we should requires this much code just to mount some disk. I can't imagine that is what everybody does. @loaj84 

We ought to find some better abstraction. Maybe we require that infrastructure presets put in a ready to use udev rule into the `infrastructure.json`?  And if Azure needs to do something extra it can still do it in a script (ideally without having to be aware of "ubuntu" users). 
Alternatively, the installation preset should write the primary user name into some configuration that is accesssible to hook scripts via `config.sh` so that we don't have to hardcode that there.
In any case, the less code, less complicated and more robust is desireable.

@wizenink @migroble @moben  ideas?

loaj84: In AWS, the EBS volume serial is known at plan time, so the udev rule can be fully resolved in Terraform. I believe the same came be achieved in Exoscale. But in Azure, disk identity (LUN → device → udev property) can only be resolved at runtime on the VM, which is why the script handles discovery and probing. Open to moving udev rule generation into Terraform where possible (AWS), but Azure will inherently need a runtime resolution step.

Gregory-Bull-Exasol: the ssh in `outputs.tf` is also tied to the ubuntu user.

Are we hoping these infrastructure presets can be independent of the username? Or are we trying to be independent of the OS?

wizenink: Maybe a cleaner way to do this would be to make `hostDatadisk` the single source of truth for all providers.

For AWS, `node.json` could carry a persistent by-id path derived from the EBS volume id:
```
hostDatadisk = "/dev/disk/by-id/nvme-Amazon_Elastic_Block_Store_${replace(aws_ebs_volume.data_disks[n.name].id, "-", "")}"
```

For azure, that is the natural path: We have an stable contract as the persistent azure path `/desk/disk/azure/data/by-lun/0`, so we could pass that path to `hostDatadisk` instead of resolving `LUN->udev->alias` at runtime.

For exoscale, I would assume that there is also a way to do this.

Then the script would be provider-agnostic, and just wait for  `hostDatadisk` to be accessible instead of injecting `udev` rules.

nh-exa: We've decided to postpone any attempts to remove "ubuntu" dependencies for now. When we decide to support more distributions, we pick up that topic again.

@wizenink Logesh appears to believe that the Azure-specific race condition problem that you've seen earlier remains regardless about how smart we get about udev rules. I can't judge that differently at the moment, but assuming that is true, we have a significant chunk of code that needs to executed for Azure and that pretty much forces us to inject it via the script hooks. I believe.

PR #34, opened 2026-03-12
feat: add Azure deployment support for Exasol Personal

loaj84:
## Description

Adds Azure infrastructure preset support for Exasol Personal and aligns the Azure preset with the existing launcher contract used by AWS.

## Related Issue

[SPOT-28929](https://exasol.atlassian.net/browse/SPOT-28929)

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Added Azure deployment preset support for provisioning Exasol Personal on Azure VMs.
- Make disk setup in prepareExasol.sh cloud-agnostic using unified hostDatadiskMatch contract
- Add VM size validation (min 4 vCPUs / 16 GB RAM) and region availability checks
- Read Azure region from ~/.azure/config, aligned with AWS behavior
- Add owner tracking via Azure AD user principal name
- Align cloud-init compression with AWS (gzip + base64)

## Testing

Manual testing completed successfully for single- and multi-node deployments. All lifecycle deployment commands executed without issues.

### Test Details

Deployed azure deployment manually using launcher CLI.

## Checklist

- [ ] Added/updated tests
- [ ] All tests pass locally

[SPOT-28929]: https://exasol.atlassian.net/browse/SPOT-28929?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

## Additional Notes

Azure managed disks created without an explicit zone are regional resources i.e., they can attach to any VM in the same region regardless of zone. This differs from AWS, where EBS volumes are locked to a specific AZ and require zone pinning to prevent detachment on VM recreation.

Since Azure doesn't have this constraint, AZ selection logic is intentionally omitted. Adding it would introduce the same zone-locking limitation without solving an existing problem, plus add a slow SKU API call to every plan/apply. Cross-zone redundancy can be added as a future enhancement if needed.

**NOTE: Archive functionality for Azure is not yet implemented.**


assets/infrastructure/azure/cloudinit.tf
```diff
@@ -0,0 +1,201 @@
...
+      hostDatadiskSourceType = "azure-lun"
+      hostDatadiskSource     = 0
+    }
+  }
+}
+
+data "cloudinit_config" "cloud_config" {
+  for_each = { for node in local.nodes : node.name => node }
+
+  gzip          = false
```
wizenink: Not specifically related to this PR, but why are we disabling compression and encoding here?
Proabably we'll hit cloudinit size limits in the future

assets/infrastructure/azure/main.tf
```diff
@@ -0,0 +1,182 @@
...
+  name                 = "${local.deployment_id}-${each.key}-data"
+  location             = azurerm_resource_group.rg.location
+  resource_group_name  = azurerm_resource_group.rg.name
+  storage_account_type = var.disk_sku
+  create_option        = "Empty"
+  disk_size_gb         = var.data_volume_size
+  tags                 = local.common_tags
+}
+
+resource "azurerm_virtual_machine_data_disk_attachment" "data_disks" {
```
wizenink, line 176: We should explicitly require this step to be a dependency for the cloudinit path. Right now, VM bootstrap depends on a timing race between the execution of cloud-init and the disk attachment

assets/infrastructure/azure/variables_public.tf
```diff
@@ -0,0 +1,53 @@
...
+}
+
+variable "adminui_password" {
+  description = "Optional Admin UI password. If empty, a random password is generated"
+  type        = string
+  default     = ""
+  sensitive   = true
+}
+
+variable "s3_archive_enabled" {
```
wizenink: We should mark this as unreachable or add a warning until the post-script is hooked and working

assets/installation/ubuntu/files/opt/exasol_launcher/scripts/prepareExasol.sh
```diff
@@ -32,11 +32,86 @@ chmod +x c4
...
+final_disk_name="$(basename "${final_disk_path}")"
+disk_source_type="$(node_jq -er '.hostDatadiskSourceType')"
+disk_source_value="$(node_jq -er '.hostDatadiskSource')"
+
+case "${disk_source_type}" in
+  aws-ebs-volume-id)
+    data_volume_id="${disk_source_value//-/}"
+    cat <<EOF | tee /etc/udev/rules.d/90-exasol.rules
+SUBSYSTEM=="block", ENV{ID_SERIAL_SHORT}=="${data_volume_id}", OWNER="ubuntu", MODE="0660"
+SUBSYSTEM=="block", ENV{ID_SERIAL_SHORT}=="${data_volume_id}", SYMLINK+="${final_disk_name}", MODE="0660"
```
wizenink: Maybe we should do this in cloudinit.tf and pass the output result to this script injecting ENV, so to not further increase the size of the cloudinit user-data

assets/infrastructure/azure/main.tf
nh-exa: good point! isn't that an issue for AWS too?

assets/infrastructure/azure/README.md
```diff
@@ -0,0 +1,112 @@
...
+   - writes deployment metadata and node metadata under `/etc/exasol_launcher/`
+   - installs the launcher scripts and systemd units
+   - prepares the Azure data disk and exposes it as `/dev/exasol_data_01`
+3. Node initialization:
+   - systemd runs the shared preparation and installation workflow
+   - Exasol is installed using the common disk alias `/dev/exasol_data_01`
+4. Local artifacts:
+   - `deployment.json` - deployment summary
+   - `secrets.json` - generated credentials
+   - `<deployment_id>.pem` - SSH private key
```
nh-exa: I believe that should be `node_access.pem` now in v1.3.0.

assets/infrastructure/azure/security.tf
```diff
@@ -0,0 +1,82 @@
+# ===============
```
nh-exa: why this header?

assets/infrastructure/azure/cloudinit.tf
loaj84: Addressed.

assets/infrastructure/azure/variables_public.tf
loaj84: archive backup is not yet taken care for azure. I have commented out the code as of now.

assets/infrastructure/azure/README.md
loaj84: Addressed.

assets/infrastructure/azure/security.tf
loaj84: Removed the headers in all files wherever applicable.

assets/infrastructure/azure/main.tf
loaj84: **Azure:**
A Terraform-level dependency isn't possible here due to how Azure's VM lifecycle works:

custom_data (cloud-init) is embedded in the `azurerm_linux_virtual_machine` resource itself, it starts executing the moment the VM boots. `azurerm_virtual_machine_data_disk_attachment` is a separate Terraform resource that can only run after the VM exists

Even if we add depends_on between these resources, that only controls Terraform's execution order, it cannot pause cloud-init inside the VM

The newer azurerm_linux_virtual_machine resource does not support inline data disk definitions (unlike the deprecated azurerm_virtual_machine which had storage_data_disk blocks). So a separate attachment is the only option, making the race inherent to Azure's Terraform provider design.

To handle this, prepareExasol.sh uses a 30-iteration discovery loop with 1s sleep between attempts, probing LUN-based symlinks (`/dev/disk/azure/scsi1/lun0, /dev/disk/azure/data/by-lun/0`) until the disk appears. In practice, disk attachment completes within 2-3 seconds.

**AWS:**
This is not an issue on AWS, the EBS volume ID is known at plan time, so the udev rule is written before the disk arrives and matches automatically once the volume attaches. No retry needed.

assets/installation/ubuntu/files/opt/exasol_launcher/scripts/prepareExasol.sh
loaj84: Addressed. The cloud-specific case statements have been replaced with a unified `hostDatadiskMatch` contract. Each provider's `cloudinit.tf` now injects either `udevKey/udevValue` (AWS, deterministic at plan time) or `discoveryPaths` (Azure, runtime probe), and the script handles both generically.

loaj84:
@nh-exa 

Currently, Azure region resolution follows this fallback chain:

- location value from ~/.azure/config (similar to how AWS inherits region from ~/.aws/config)
- Falls back to westeurope as a hardcoded default

In AWS, the Terraform provider natively reads ~/.aws/config for the region (no custom logic needed). Azure's Terraform provider (azurerm) does not offer this behavior, so we read ~/.azure/config ourselves using file() + regex().

The hardcoded westeurope fallback is a pragmatic choice for now, but it could silently deploy infrastructure in an unexpected region for users outside Europe. A better approach would be to fail with a clear error message like:

"Azure region not configured. Set it via az configure --defaults location=<region>."

This helps in avoiding accidental cross-region deployments. Open to suggestions and discussions.

loaj84:
**Note on Azure Key Vault (commented out in ssh.tf):**

In AWS, the SSH private key is stored in AWS Systems Manager (SSM) Parameter Store as an encrypted SecureString parameter at /<deployment_id>/ssh_private_key. This ensures the key is persisted in the cloud with encryption at rest via KMS, providing a recovery path if the local PEM file is lost and an audit trail of secret access.

The Azure equivalent which is the Azure Key Vault is implemented but commented out for now. It would store the SSH private key as a Key Vault secret with RBAC-based access control. However, creating the required azurerm_role_assignment (Key Vault Secrets Officer) failed with a 403 error because the deploying identity lacks Microsoft.Authorization/roleAssignments/write permission, which requires the Owner or User Access Administrator role on the subscription. I will go ahead and raise an ITM for that.

This will be enabled once the required RBAC permissions are granted. Until then, SSH keys are stored only locally as node_access.pem.

nh-exa:
> @nh-exa
> 
> Currently, Azure region resolution follows this fallback chain:
> 
> * location value from ~/.azure/config (similar to how AWS inherits region from ~/.aws/config)
> * Falls back to westeurope as a hardcoded default
> 
> In AWS, the Terraform provider natively reads ~/.aws/config for the region (no custom logic needed). Azure's Terraform provider (azurerm) does not offer this behavior, so we read ~/.azure/config ourselves using file() + regex().
> 
> The hardcoded westeurope fallback is a pragmatic choice for now, but it could silently deploy infrastructure in an unexpected region for users outside Europe. A better approach would be to fail with a clear error message like:
> 
> "Azure region not configured. Set it via az configure --defaults location=."
> 
> This helps in avoiding accidental cross-region deployments. Open to suggestions and discussions.

I can't say I like us accessing and parsing "~/.azure/config"

Why wouldn't we just add a "location" variable to "variables_public.tf" and give it a default? Or no default? We could ask @juergen-albertsen-exasol  if he likes it better if people need to provide this or risk it being in the wrong region.
The problem I see is that if you have a default, it gets deployed "somewhere" and the costs and latency might be affected negatively.

loaj84:
> > @nh-exa
> > Currently, Azure region resolution follows this fallback chain:
> > 
> > * location value from ~/.azure/config (similar to how AWS inherits region from ~/.aws/config)
> > * Falls back to westeurope as a hardcoded default
> > 
> > In AWS, the Terraform provider natively reads ~/.aws/config for the region (no custom logic needed). Azure's Terraform provider (azurerm) does not offer this behavior, so we read ~/.azure/config ourselves using file() + regex().
> > The hardcoded westeurope fallback is a pragmatic choice for now, but it could silently deploy infrastructure in an unexpected region for users outside Europe. A better approach would be to fail with a clear error message like:
> > "Azure region not configured. Set it via az configure --defaults location=."
> > This helps in avoiding accidental cross-region deployments. Open to suggestions and discussions.
> 
> I can't say I like us accessing and parsing "~/.azure/config"
> 
> Why wouldn't we just add a "location" variable to "variables_public.tf" and give it a default? Or no default? We could ask @juergen-albertsen-exasol if he likes it better if people need to provide this or risk it being in the wrong region. The problem I see is that if you have a default, it gets deployed "somewhere" and the costs and latency might be affected negatively.

Agreed. I'll remove the config file parsing and move location to variables_public.tf as a required variable with no default. This forces users to explicitly specify their region, avoiding accidental deployments in an unintended region with unexpected cost/latency. I will check with Jurgen also on the same.

loaj84:
> > > @nh-exa
> > > Currently, Azure region resolution follows this fallback chain:
> > > 
> > > * location value from ~/.azure/config (similar to how AWS inherits region from ~/.aws/config)
> > > * Falls back to westeurope as a hardcoded default
> > > 
> > > In AWS, the Terraform provider natively reads ~/.aws/config for the region (no custom logic needed). Azure's Terraform provider (azurerm) does not offer this behavior, so we read ~/.azure/config ourselves using file() + regex().
> > > The hardcoded westeurope fallback is a pragmatic choice for now, but it could silently deploy infrastructure in an unexpected region for users outside Europe. A better approach would be to fail with a clear error message like:
> > > "Azure region not configured. Set it via az configure --defaults location=."
> > > This helps in avoiding accidental cross-region deployments. Open to suggestions and discussions.
> > 
> > 
> > I can't say I like us accessing and parsing "~/.azure/config"
> > Why wouldn't we just add a "location" variable to "variables_public.tf" and give it a default? Or no default? We could ask @juergen-albertsen-exasol if he likes it better if people need to provide this or risk it being in the wrong region. The problem I see is that if you have a default, it gets deployed "somewhere" and the costs and latency might be affected negatively.
> 
> Agreed. I'll remove the config file parsing and move location to variables_public.tf as a required variable with no default. This forces users to explicitly specify their region, avoiding accidental deployments in an unintended region with unexpected cost/latency. I will check with Jurgen also on the same.

Approach `Location parameter required as part of launcher command` has been implemented and changes are pushed.

assets/infrastructure/azure/main.tf
loaj84: udev's built-in device event matching has been used to eliminate the runtime disk discovery loop on Azure. The full udev match clause is now resolved at Terraform plan time using the deterministic LUN-based symlink path. The rule is written upfront and udev applies it automatically when the disk appears, removing the 30-second polling loop and simplifying prepareExasol.sh into a provider-agnostic script.

assets/infrastructure/azure/README.md
```diff
@@ -0,0 +1,112 @@
...
+- The default image is Canonical Ubuntu 22.04 LTS Gen2 (`latest`).
+- SSH access is configured with a generated RSA key pair and password authentication is disabled.
+
+### Storage
+- Each node gets a separate OS disk and a separate managed data disk.
+- - Both disks use the Azure managed disk SKU configured by `disk_sku` (default: `StandardSSD_LRS`).
+- The data disk is attached at LUN `0`.
+- The node metadata exposes a provider-neutral final disk alias `/dev/exasol_data_01`.
+- During node preparation, the installer resolves the Azure disk from its LUN and creates the `/dev/exasol_data_01` alias via `udev`, so Exasol can use the same final disk path across providers.
+- Remote archive integration is not currently implemented in this Azure preset.
```
nh-exa, line 24: if that is true, you'll want to remove "assets/infrastructure/azure/files/opt/exasol_launcher/scripts/azure_registerArchiveVolume.sh" from this PR

assets/infrastructure/azure/ssh.tf
```diff
@@ -0,0 +1,40 @@
...
+  rsa_bits  = 4096
+}
+
+data "azurerm_client_config" "current" {}
+
+data "azuread_user" "current" {
+  object_id = data.azurerm_client_config.current.object_id
+}
+
+# resource "azurerm_key_vault" "deployment" {
```
nh-exa: what is all of this? remove it if it's not needed please

assets/infrastructure/azure/rbac-role.minimal.json
nh-exa, line 1: I suspect that "IAM" policy is not a good file name here.

nh-exa (changes requested):

assets/infrastructure/azure/README.md
loaj84: addressed

assets/infrastructure/azure/ssh.tf
loaj84: addressed

assets/infrastructure/azure/rbac-role.minimal.json
loaj84: addressed

nh-exa (approved):

PR #42, opened 2026-03-18
feat: Implement exasol shell container to open an interactive shell into cos…

wizenink:
## Description

Added a `exasol shell container` command to open an interactive shell into the cos container, for easy access into database administration.
Moved `exasol diag shell` to  `exasol shell host` and left the original route intact with a deprecation warning

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Added and documented the `exasol diag cos` command
- Moved `exasol diag shell` to  `exasol shell host`
- Added utility methods for executing an interactive shell through an SSH connection
- Added e2e tests for the `exasol shell container` command

## Testing

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

### DIFF
--- a/internal/deploy/shell.go
+++ b/internal/deploy/shell.go
@@ -16,24 +16,43 @@ import (
 
 var ErrNoNodesFound = errors.New("no nodes found in the active deployment")
 
-// Start an interactive shell using stdin stdout & stderr.
-func Shell(ctx context.Context, deploymentDir string, selectedNode string) error {
+// OpenHostShell starts an interactive shell using stdin stdout & stderr.
+func OpenHostShell(ctx context.Context, deploymentDir string, selectedNode string) error {
 	return withDeploymentSharedLock(ctx, deploymentDir, func(dir string) error {
-		return shellUnsafe(ctx, dir, selectedNode)
+		sshRemote, err := sshRemoteForNodeUnsafe(dir, selectedNode)
+		if err != nil {
+			return err
+		}
+
+		return sshRemote.Shell(ctx, os.Stdout, os.Stderr)
 	})
 }
 
-func shellUnsafe(ctx context.Context, deploymentDir string, selectedNode string) error {
+// OpenCOSShell opens an interactive COS session via the access node (n11).
+func OpenCOSShell(ctx context.Context, deploymentDir string) error {
+	return withDeploymentSharedLock(ctx, deploymentDir, func(dir string) error {
+		sshRemote, err := sshRemoteForNodeUnsafe(dir, "n11")
+		if err != nil {
+			return err
+		}
+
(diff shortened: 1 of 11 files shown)


internal/deploy/shell.go
```diff
@@ -16,32 +16,53 @@ import (
...
+		if err != nil {
+			return err
+		}
+
+		return sshRemote.Shell(ctx, os.Stdout, os.Stderr)
+	})
+}
+
+// COS opens an interactive COS session via the access node (n11).
+func COS(ctx context.Context, deploymentDir string) error {
```
nh-exa: I know "Shell" started that trend but I'd prefer if our methods have verbs. "openHostShell" and "openCOSShell"?

internal/deploy/shell.go
```diff
@@ -16,32 +16,53 @@ import (
...
+	return withDeploymentSharedLock(ctx, deploymentDir, func(dir string) error {
+		sshRemote, err := sshRemoteForNodeUnsafe(dir, "n11")
+		if err != nil {
+			return err
+		}
+
+		cosCommand := `/bin/bash -lc '` +
```
nh-exa: I'm quite a bit queasy about including this bash code into the launcher here. The reason is that I feel the launcher should make less assumptions and integrate less with internals of the Exasol installation (or cloud providers too). The installation preset effectively decides where the c4 binary sits and how to a find the play id. if we can keep c4 out of here that would be best.
How about you implement that detail in the installation preset and have the launcher just call "connect_cos.sh"?

Though of course, the trouble at the moment is that any addition code to the cloud-config/init might break the AWS deployment ...

nh-exa (changes requested):

internal/deploy/shell.go
nh-exa: If we think ahead a bit, we could even argue that "cos" isn't a concept that should be made known here either. What we are really doing is connecting to the container on the host in which the DB is running in. It just happens to be currently be part of COS. That might change as we are going to be debundling a few things there this year (hopefully).

nh-exa: `exasol shell host` and `exasol shell container`? @juergen-albertsen-exasol

wizenink: Makes sense to me. Moving the commands from `diag` to `shell` is also a good move IMHO, since they are essentially administration tools not only used for diagnostics.

wizenink: Regarding the cloud-init file size, the `connectCos.sh` script would only take up 279bytes uncompressed, but I see the point of being extra careful when adding things to cloud-init. We can maybe move some other things into a downloadable tarball, we'll talk

nh-exa (approved):

PR #59, opened 2026-04-02
Gb.exoscale cleanup tool support

Gregory-Bull-Exasol:
moved to https://github.com/exasol/exasol-personal/pull/86


tools/cleanup/internal/exoscale/handlers.go
```diff
@@ -0,0 +1,311 @@
...
+	"context"
+	"errors"
+	"fmt"
+	"strings"
+	"time"
+
+	awssdk "github.com/aws/aws-sdk-go-v2/aws"
+	awsconfig "github.com/aws/aws-sdk-go-v2/config"
+	"github.com/aws/aws-sdk-go-v2/service/s3"
+	egoscale "github.com/exoscale/egoscale/v2"
```
nh-exa: > v1 and v2 of egoscale are deprecated, please use [v3](https://pkg.go.dev/github.com/exoscale/egoscale/v3)

See: https://github.com/exoscale/egoscale

tools/cleanup/go.mod
```diff
@@ -11,10 +11,40 @@ require (
...
 	github.com/aws/aws-sdk-go-v2/service/sts v1.41.3
-	github.com/spf13/cobra v1.8.0
+	github.com/exoscale/egoscale v0.102.4
+	github.com/spf13/cobra v1.10.2
 )
 
+tool github.com/golangci/golangci-lint/v2/cmd/golangci-lint
+
 require (
+	4d63.com/gocheckcompilerdirectives v1.3.0 // indirect
```
nh-exa, line 21: How did we end up with a bazillion new 3rd party dependencies? I think it is mostly the linter? I've check a few and almost all are something transitive from golang-ci.

What happened to your disdain for linters? :D
For this thing, I had decided not to use a linter to make iteration cycles with LLMs faster.

Also, since there is a massive supply chain attack almost daily now, I felt like reducing our dependencies might be good idea.

nh-exa: Ultimately, I think I'd rather build the linter as binary once and just invoke it here instead of building it again and again from source in every repo. Out of scope here obviously.

nh-exa (changes requested):
I've tested it without authentication to either AWS or exoscale.
No matter if i provide the AWS region or a exoscale zone, tt just reports "0 deployments" and i'd rather have it tell me that authentication for a certain provider failed and have the tool exit non-zero too. That is critical if we want to use this in automation.
We also need somehow be able to select a list of providers. Maybe simply by adding this option to the CLI:
[--provider all/aws/exoscale] - default being "all"

nh-exa (dismissed):

loaj84 (dismissed):

loaj84 (dismissed):

nh-exa:
Some if not all commit should be squashed.

Gregory-Bull-Exasol:
Moved to PR https://github.com/exasol/exasol-personal/pull/86
github automated merge made the history hard to edit.

PR #63, opened 2026-04-06
Make paths relative to deployment directory

migroble:
## Description

Some paths in the configuration files that should be relative to the deployment directory were being stored as absolute, making it so the deployment directory couldn't be moved once it was created.

This issue affected the infrastructure and installation directories and the paths to the SSH key files.

## Related Issue

[SPOT-29898](https://exasol.atlassian.net/browse/SPOT-29898)

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- Removed `infrastructure_artifact_dir` from the tfvars file. It is now passed directly as an argument to tofu.
- Added `DeploymentPath` type to enforce paths being relative to the deployment directory on the type level.

## Testing

### Test Details

To manually test this change a deployment was initialized and deployed and then, the command `grep -r "$PWD"` was executed in the deployment directory. The only file that should show up is `deployment.log`.

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

[SPOT-29898]: https://exasol.atlassian.net/browse/SPOT-29898?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/tofu/operations.go
+++ b/internal/tofu/operations.go
@@ -114,6 +114,7 @@ func ApplyPlan(
 func ApplyAction(
 	ctx context.Context,
 	cfg Config,
+	deploymentDir string,
 	action string,
 	out, outErr io.Writer,
 ) error {
@@ -123,7 +124,7 @@ func ApplyAction(
 
 	applyOpts := ApplyOptions{
 		// Cannot use plan file
-		VarArgs: []string{action},
+		VarArgs: []string{action, "infrastructure_artifact_dir=" + deploymentDir},
 		// Must use Vars file
 		VarsFilePath:  cfg.VarsOutputFile(),
 		StateFilePath: cfg.StateFile(),
--- a/internal/util/util.go
+++ b/internal/util/util.go
@@ -72,6 +72,40 @@ func AbsPathNoFail(path string) string {
 	return absPath
 }
 
+type DeploymentPath struct {
+	path string
+}
+
+func NewDeploymentPath(path string, deploymentDir string) DeploymentPath {
+	if filepath.IsAbs(path) {
+		relPath, err := filepath.Rel(deploymentDir, path)
+		if err != nil {
+			panic("Path not in deployment directory")
+		}
+
+		return DeploymentPath{relPath}
+	}
+
+	return DeploymentPath{path}
+}
+
+func (o *DeploymentPath) Abs(deploymentDir string) string {
+	return filepath.Join(deploymentDir, o.path)
+}
+
+func (o *DeploymentPath) String() string {
+	return o.path
+}
+
+func (o *DeploymentPath) MarshalText() ([]byte, error) {
+	return []byte(o.path), nil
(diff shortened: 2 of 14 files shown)


internal/tofu/operations.go
```diff
@@ -123,7 +124,7 @@ func ApplyAction(
 
 	applyOpts := ApplyOptions{
 		// Cannot use plan file
-		VarArgs: []string{action},
+		VarArgs: []string{action, "infrastructure_artifact_dir=" + deploymentDir},
```
nh-exa, line 127: I haven't quite understood why you would pass it like that instead of with the vars file.
Is it because "deploymentDir" is still absolute here? If that is the case, why not write a relative path to the vars file instead?

internal/util/util.go
```diff
@@ -72,6 +72,40 @@ func AbsPathNoFail(path string) string {
 	return absPath
 }
 
+type DeploymentPath struct {
+	path string
+}
+
+func NewDeploymentPath(path string, deploymentDir string) DeploymentPath {
```
nh-exa, line 79: I don't understand this type and function. If `path` is not absolute `deploymentDir` is not used at all?
What is a DeploymentPath anyway?  

I'd like to see some unit tests for this to help me understand.
Also, I suspect that "DeploymentPath" is not a good name for this type. It appears to be a utility for managing relatives and not to be specific for the "deployment directory"?

nh-exa (changes requested):
Overall, i got to wonder if this is the simplest solution to this problem.
What we just made the  directories being passed to the preset in TFvars relative just when it's written?

internal/tofu/operations.go
migroble: Since `infrastructure_artifact_dir` is always an absolute path to the deployment directory, if we store it in the vars file, we will be storing the absolute path to the deployment directory somewhere inside the deployment directory, which is what causes things to break if we move it.

Another way to look at it is that the `infrastructure_artifact_dir` is an inherently runtime-only variable and can change from execution to execution, so we shouldn't cache it anywhere.

internal/util/util.go
migroble: This type doesn't really do anything other than enforcing that the path it stores is relative (to the deployment directory) and gives a type-level hint of what the value is instead of just being a generic string with some naming convention. This pattern is pretty common in Rust where it's called "NewType pattern", I'm not sure if it's very idiomatic in Go though.

I agree the name could be made more generic (something like RelPath), however it only makes sense for paths within the deployment directory since those are the only paths that must be relative. I'm open for suggestions on this one -- or if we want to scrap the entire concept.

nh-exa:
Replaced by https://github.com/exasol/exasol-personal/pull/83

PR #73, opened 2026-04-09
feat: add vm disk image building util for exasol nano

Gregory-Bull-Exasol:
Workflow automation for building single-container podman virtual machines for execution on mac (and maybe windows)

- Downloads a cloudinit capable alpine linux image
- Expands it to make space for changes
- Uses cloudinit to
    - Install podman (and jq)
    - Prepare for mounted folders from the host using virtiofs (or smb on windows hosts)
    - Setup ssh key loading from that mounted folder
    - Setup podman container loading from that mounted folder
    - Preload a container into the VM
- Shrink the disk image back down to a minimum size
- Make mac vfkit (and windows hyper-v) deployment packages.

See buildvm/README.md

Uses QEMU for vm operations on linux. Supports building x86 and arm64 vm images.


tools/buildvm/cloud-init/scripts/load-shared-container.sh
```diff
@@ -0,0 +1,267 @@
...
+
+# Constants
+CONTAINER_NAME="container"
+STATE_FILE="/var/lib/container-state.sha256"
+
+# Parse mounts from manifest (will be set later if manifest exists)
+VOLUME_FLAGS=""
+
+# Check if manifest exists, fall back to stored manifest if available
+SKIP_LOAD=false
```
loaj84, line 41: Lines `41-54` and `56-71` are duplicate blocks that both check if the manifest file exists and fall back to the stored manifest. The first block sets `SKIP_LOAD`, the second sets `HAS_MANIFEST`, but then `SKIP_LOAD` is overwritten again at line `123`. The first block (lines 41-54) can be removed entirely as only the `HAS_MANIFEST` block is used downstream.

tools/buildvm/cloud-init/scripts/load-shared-container.sh
```diff
@@ -0,0 +1,267 @@
...
+# Container runtime log file
+CONTAINER_LOG_FILE="$LOG_DIR/container-runtime-$(date +%Y%m%d-%H%M%S).log"
+
+# Run the container with args (as root with proper cgroup2 support)
+# NOTE: No --cpus or --memory limits are set intentionally.
+# This allows the container to automatically use all VM resources.
+# If VM resources are increased (e.g., more CPUs or RAM), the container
+# will automatically have access to them without configuration changes.
+log_msg "Starting container (ports: $CONTAINER_PORTS)..."
+if podman run -d \
```
loaj84, line 252: podman run here doesn't support passing --shm-size. Exasol-nano requires --shm-size 1g to start. Suggest adding a shmSize field to the container manifest and passing it through here as we are planning to have podman running in the VM

tools/buildvm/wget-log
```diff
@@ -0,0 +1,11 @@
+--2026-04-10 19:31:10--  https://dl-cdn.alpinelinux.org/alpine/v3.23/releases/cloud/nocloud_alpine-3.23.3-aarch64-uefi-cloudinit-r0.qcow2
+Resolving dl-cdn.alpinelinux.org (dl-cdn.alpinelinux.org)... 146.75.122.132
```
loaj84, line 2: do we need this file here?

loaj84 (changes requested):

tools/buildvm/scripts/start-vm.sh.backup
```diff
@@ -0,0 +1,78 @@
+#!/usr/bin/env bash
```
loaj84, line 1: hope this file also can be removed.

nh-exa:
I think this is no longer needed in this repo. Closing.

PR #74, opened 2026-04-10
feat: accept multiline queries by executing on semicolon

wizenink:
## Description

Default `connect` shell behavior now executes SQL only after a terminating semicolon (;), which supports multiline/pasted SQL statements. A compatibility flag allows reverting to line-by-line execution.

## Related Issue

Fixes #

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Added shell execution options with semicolon mode support
- Implemented statement buffering in semicolon mode
- Wired connect to pass semicolon execution mode
- Added CLI flag `--execute-on-semicolon` with a `true` default

## Testing

## Checklist

- [] Updated documentation (if applicable)
- [x] Added/updated tests
- [x] All tests pass locally

### DIFF
--- a/internal/connect/shell.go
+++ b/internal/connect/shell.go
@@ -21,13 +21,23 @@ const exitCommand = "exit"
 // ProcessInputFunc defines a way to process a shell input.
 type ProcessInputFunc func(input string) error
 
-type shell struct {
-	lineReader   types.LineReader
-	processInput ProcessInputFunc
+type ShellOpts struct {
+	ExecuteOnSemicolon bool
 }
 
-func newShell(lineReader types.LineReader, processInput ProcessInputFunc) *shell {
-	return &shell{lineReader, processInput}
+type shell struct {
+	lineReader          types.LineReader
+	processInput        ProcessInputFunc
+	executeOnSemicolon  bool
+	pendingStatementBuf string
+}
+
+func newShell(lineReader types.LineReader, processInput ProcessInputFunc, opts ShellOpts) *shell {
+	return &shell{
+		lineReader:         lineReader,
+		processInput:       processInput,
+		executeOnSemicolon: opts.ExecuteOnSemicolon,
+	}
 }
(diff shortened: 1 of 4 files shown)


andymaclean:
Hi, from my reading of this you are accepting all semicolons which are not inside single quotes as terminators.  Normally you also want to ignore double quotes (this is quoted identifier syntax in SQL and allowing a column named ';' to break your SQL into two is a security problem).  Also one normally excludes semicolons in comments.  At a minimum you would want SQL style comments, which are -- up to the end of the line.  Often I like to also do bash and C/C++ style comments although those are a bit more optional and it depends on the tool you use.  You should probably check which comment styles the Exasol SQL compiler understands and support those.

internal/connect/shell.go
```diff
@@ -49,19 +59,76 @@ func (sh *shell) run() error {
...
+	if sh.pendingStatementBuf != "" {
+		sh.pendingStatementBuf += "\n"
+	}
+	sh.pendingStatementBuf += line
+
+	statements, remainder := splitSemicolonTerminatedStatements(sh.pendingStatementBuf)
+	sh.pendingStatementBuf = remainder
+
+	for _, statement := range statements {
+		if err := sh.processInput(strings.TrimSpace(statement)); err != nil {
```
loaj84, line 90: When multiple semicolon-terminated statements appear on the same line (e.g., `SELECT 1;INVALID SQL;SELECT 2;`), the loop defined here returns on first error, silently dropping remaining statements. This is an edge but it will be better if we consider logging the error and continuing the loop instead of returning early. This works fine if the same statements are in separate lines.

loaj84 (dismissed):
Just left a minor comment on the error handling in the semicolon loop. Rest of my feedback is already covered in Andy's review.

nh-exa (approved):
I think this is fine.

PR #83, opened 2026-04-16
fix: make deployment paths relative and launcher-owned

nh-exa:
## Description

Make deployment directories movable by storing launcher-managed paths as relative paths and keeping the deployment layout contract in the launcher.

Replaces: https://github.com/exasol/exasol-personal/pull/63

## Related Issue

Fixes SPOT-29898

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- Moved the deployment path contract into launcher-owned helpers and switched injected infrastructure paths to relative values
- Updated infrastructure presets to write relative `keyFile` paths while keeping SSH command rendering in the presets
- Backward compatibility: existing deployments initialized by older launchers continue to work at their original location
- Limitation: deployments initialized before this change are not retroactively made movable if they already contain absolute persisted paths

## Testing

### Test Details

Validated the change with focused unit tests around deployment initialization and node-details path handling, then ran the full test suite. Manually tested with AWS deployment.

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

## Additional Notes

This change keeps the deployment directory layout as a launcher-owned contract instead of having presets infer each other's location. New deployments store relative paths, while legacy deployments with absolute `keyFile` paths remain supported.


assets/infrastructure/aws/outputs.tf
```diff
@@ -29,9 +31,9 @@ locals {
         ssh = {
           username = "ubuntu"
           keyName  = local.ssh_key_name
-          keyFile  = local.key_file_path
+          keyFile  = local.key_file_relative_path
           port     = "22"
-          command  = "ssh -i ${local.key_file_path} ubuntu@${node.public_dns} -p 22"
+          command  = "ssh -i ${local.key_file_relative_path} ubuntu@${node.public_dns} -p 22"
```
wizenink, line 36: We should use `ResolveDeploymentPath` here aswell, otherwise copypasting the ssh command would only work if the user is on the deployment dir

assets/infrastructure/azure/outputs.tf
```diff
@@ -23,9 +25,9 @@ locals {
         ssh = {
           username = "ubuntu"
           keyName  = local.ssh_key_name
-          keyFile  = local.key_file_path
+          keyFile  = local.key_file_relative_path
           port     = "22"
-          command  = "ssh -i ${local.key_file_path} ubuntu@${azurerm_public_ip.nodes[k].ip_address} -p 22"
+          command  = "ssh -i ${local.key_file_relative_path} ubuntu@${azurerm_public_ip.nodes[k].ip_address} -p 22"
```
wizenink, line 30: We should use `ResolveDeploymentPath` here aswell, otherwise copypasting the ssh command would only work if the user is on the deployment dir

assets/infrastructure/exoscale/outputs.tf
```diff
@@ -23,9 +25,9 @@ locals {
         ssh = {
           username = "ubuntu"
           keyName  = local.ssh_key_name
-          keyFile  = local.key_file_path
+          keyFile  = local.key_file_relative_path
           port     = "22"
-          command  = "ssh -i ${local.key_file_path} ubuntu@${exoscale_compute_instance.nodes[node_config.name].public_ip_address} -p 22"
+          command  = "ssh -i ${local.key_file_relative_path} ubuntu@${exoscale_compute_instance.nodes[node_config.name].public_ip_address} -p 22"
```
wizenink, line 30: We should use `ResolveDeploymentPath` here aswell, otherwise copypasting the ssh command would only work if the user is on the deployment dir

wizenink (changes requested):
The launcher owns `config.NodeAccessKeyFileName = "node_access.pem"`, and each TF preset hardcodes `key_file_name = "node_access.pem"`. If this changes, an stale preset would override it and leave the launcher unnoticed. Maybe we could pass the go owned string into the TF presets.

`ResolveDeploymentPath` could be expanded to resolve via `NodeDetails.Resolve(deploymentDir) -> ResolvedNodeDetails` which, even if currently would only resolve the ssh dir, can be expanded to resolve every path field, and make impossible for devs to forget to call the resolver routine.

assets/infrastructure/aws/outputs.tf
nh-exa: We envisioned that at some point the launcher wouldn't be aware of whether ssh is used or e.g. some cloud-specific things like AWS SSM. So presets would have to provide the means of connecting to the instances somehow. That is why we have that "command" field in the outputs.

if the entire command is generated, obviously no assumption can be made about about from where this will be called. Effectively that means these commands always have to be called from the deployment directory.

The Exasol Personal launcher would need to change the pwd the ssh command to that. And if the we are using an SSH library, the connection is assembled differently anyway and it can just pick up the keyfile.

In any case, we can't use `ResolveDeploymentPath` in Terraform because it is not defined there. Maybe once we have a shared tofu library preset.

nh-exa:
> The launcher owns `config.NodeAccessKeyFileName = "node_access.pem"`, and each TF preset hardcodes `key_file_name = "node_access.pem"`. If this changes, an stale preset would override it and leave the launcher unnoticed. Maybe we could pass the go owned string into the TF presets.

Yeah, that filename is hard-coded at the moment. As are some more like "deployment.json".
I'm not sure if I want to change that just for this file now. These hardcoded file names are documented in the `doc/presets.md#L41`

> `ResolveDeploymentPath` could be expanded to resolve via `NodeDetails.Resolve(deploymentDir) -> ResolvedNodeDetails` which, even if currently would only resolve the ssh dir, can be expanded to resolve every path field, and make impossible for devs to forget to call the resolver routine.

That might be worthwhile. I'll look into it

internal/deploy/init.go
```diff
@@ -143,10 +143,12 @@ func InitDeployment(
...
-			// These values should always be part of the infra vars per contract
-			// It tell the infrastructure preset where to write deployment artifacts fot the launcher
-			infraVars["infrastructure_artifact_dir"] = deploymentDir
-			infraVars["installation_preset_dir"] = installDir
+			// These values should always be part of the infra vars per contract.
+			// They are expressed relative to the extracted infrastructure preset
+			// directory, which keeps the deployment directory movable while
+			// preserving a single launcher-owned layout SSOT.
+			infraVars["infrastructure_artifact_dir"] = config.RelativeInfrastructureArtifactDir()
+			infraVars["installation_preset_dir"] = config.RelativeInstallationPresetDir()
```
migroble: Since these values are now always relative to Tofu's workdir and constant, could we set them as the default values in the `variables_internal.tf` files and remove them from here?

internal/config/paths.go
```diff
@@ -11,4 +16,27 @@ const (
 	InfrastructureFilesDirectory = "infrastructure"
 	InstallationFilesDirectory   = "installation"
 	SharedFilesDirectory         = "."
+	NodeAccessKeyFileName        = "node_access.pem"
```
migroble: This constant appears to only be used in the tests, would it make sense to move it there instead?

internal/deploy/init.go
nh-exa: The reason why I wanted those in the first place and keep them is so that the Launcher has full control over the layout of the deployment directory and the presets don't assume a directory layout. That gives us a SSOT - which is always nice.

wizenink (dismissed):

wizenink (approved):

PR #85, opened 2026-04-20
fix(exoscale): correct iam example policies

Gregory-Bull-Exasol:


nh-exa (approved):

nh-exa:
May I ask what was wrong here in the first place?

Gregory-Bull-Exasol:
> May I ask what was wrong here in the first place?

They were not tested

PR #87, opened 2026-04-21
feat: embed compressed tofu instead of uncompressed binary

wizenink:
## Related Issue

Fixes #

## Type of Change

- [ ] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

-
-
-

## Testing

## Checklist

- [ ] Added/updated tests
- [ ] All tests pass locally

### DIFF
--- a/tools/downloadtofu/main.go
+++ b/tools/downloadtofu/main.go
@@ -4,8 +4,7 @@
 package main
 
 import (
-	"archive/tar"
-	"compress/gzip"
+	"bytes"
 	"context"
 	"flag"
 	"fmt"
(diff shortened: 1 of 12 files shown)


tools/downloadtofu/main.go
```diff
@@ -63,72 +63,23 @@ func downloadTofu() error {
...
-		fmt.Println("open tofu is already download") // nolint: revive,forbidigo
-	}
-
-	compressedFile, err := os.Open(compressedFilePath)
-	if err != nil {
-		return fmt.Errorf("failed to open compressed file: %w", err)
+	legacyArchivePath := path.Join(*destDir, tofuReleaseFileName(*goos, *goarch))
+	if err := migrateLegacyArchivePath(legacyArchivePath, compressedFilePath); err != nil {
+		return err
 	}
```
migroble: Aren't we downloading to the correct path already? Is this migration necessary?

wizenink: I was thinking about having deployments that were being updated, but we are not focusing on long lived deployments, so this should not be necessary, indeed.
Good catch, fixing

migroble (dismissed):

nh-exa:
Since we've decided we are not going to embed the tofu binary like that anymore soon, there is no real point in merging this, is there? I'll close it.

PR #93, opened 2026-04-28
fix: flush pending statement on EOF in semicolon mode

loaj84:
## Description

When SQL is piped into `exasol connect` (non-interactive stdin), the shell silently dropped the final statement if it lacked a trailing `;`. No error, exit code 0.

Examples that were broken:
- `echo "SELECT 1" | exasol connect`
- `exasol connect < migration.sql` where the file's last statement has no `;`
- The CI tests `test_single_query` and `test_file_import`, which pipe queries without `;`

## Related Issue

Fixes #

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- On non-interactive io.EOF, flush any buffered statement before returning, so piped or redirected input without a trailing ; no longer drops the final statement. Interactive EOF still discards it.
- ErrInterrupt (Ctrl-C) keeps its current behavior and does not execute the buffered statement.
- Extracted the repeated processInput + slog.Error handling into a small execStatement helper.
- Added unit tests covering both EOF behaviors.

## Testing

### Test Details

Unit Testing has been written to cover the code changes.

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

### DIFF
--- a/internal/connect/shell.go
+++ b/internal/connect/shell.go
@@ -23,19 +23,12 @@ type ProcessInputFunc func(input string) error
 
 type ShellOpts struct {
 	ExecuteOnSemicolon bool
-	// FlushPendingOnEOF makes the shell execute any buffered (un-`;`-terminated)
-	// statement when the input stream closes. Intended for non-interactive stdin
-	// (pipes, redirected files) so the final statement is not silently dropped.
-	// Must stay false for interactive sessions, where Ctrl-D after typing an
-	// incomplete statement should discard the buffer rather than execute it.
-	FlushPendingOnEOF bool
 }
 
 type shell struct {
 	lineReader          types.LineReader
 	processInput        ProcessInputFunc
 	executeOnSemicolon  bool
-	flushPendingOnEOF   bool
 	pendingStatementBuf string
 }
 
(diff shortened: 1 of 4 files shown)


wizenink (dismissed):
Nice catch, ty!

wizenink (dismissed):

internal/connect/shell.go
```diff
@@ -23,12 +23,19 @@ type ProcessInputFunc func(input string) error
 
 type ShellOpts struct {
 	ExecuteOnSemicolon bool
+	// FlushPendingOnEOF makes the shell execute any buffered (un-`;`-terminated)
+	// statement when the input stream closes. Intended for non-interactive stdin
+	// (pipes, redirected files) so the final statement is not silently dropped.
+	// Must stay false for interactive sessions, where Ctrl-D after typing an
+	// incomplete statement should discard the buffer rather than execute it.
```
moben: This seems strange to me, I never noticed that behavior. I just checked some other interactive prompts (bash and python) and the way it works there is that Ctrl-D only discards if the current line is empty.

See e.g. https://unix.stackexchange.com/questions/556675/how-to-follow-up-a-command-with-ctrl-d-eof

We don't necessarily need to change this but it might be nice to align with what everyone is doing here.

loaj84: Makes sense. I changed this so EOF preserves already-buffered input instead of discarding it, which feels closer to standard shell behavior.

moben (approved):

PR #96, opened 2026-05-05
feat: Add STACKIT infrastructure support

migroble:
## Description

This PR adds support for deployment Exasol Personal on STACKIT.

~~Can't test this PR at the moment due to a breaking bug in STACKIT's Terraform provider. :/~~ They fixed it!

## Related Issue

[SPOT-30098](https://exasol.atlassian.net/browse/SPOT-30098)

## Type of Change

- [X] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Create OpenTofu definitions to deploy Exasol Personal on STACKIT. 
- Add STACKIT to the tests.
- Add STACKIT to the cleanup tool.

## Testing

- [X] Manually tested the changes

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

[SPOT-30098]: https://exasol.atlassian.net/browse/SPOT-30098?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


wizenink (commented):
We should also add stackit into `README.md`, `cmd/exasol/root.go` and the deployment tests.

migroble:
Done. I thought "adding the tests" was a different task, but it was "adding the tests to CI". Updating README + adding the how-to guide *is* a different task, so I'll leave that out of this PR.

assets/infrastructure/stackit/cloudinit.tf
```diff
@@ -0,0 +1,174 @@
...
+  }
+
+  # Per-node metadata written to disk (separate from infrastructure.json for clarity).
+  node_payload_by_name = {
+    for n in local.nodes : n.name => {
+      name      = n.name
+      privateIp = n.ip
+      myId      = n.name
+
+      # TODO: Change this for StackIT.
```
nh-exa: guess that TODO is done?

assets/infrastructure/stackit/cloudinit.tf
```diff
@@ -0,0 +1,174 @@
...
+  base64_encode = false
+
+  dynamic "part" {
+    for_each = local.installation_cloudconf_files
+
+    content {
+      content_type = "text/cloud-config"
+      # Numeric prefix makes ordering/precedence explicit when debugging multipart user-data.
+      filename = "10-cloudconf-${part.value}"
+      content  = file("${local.installation_cloudconf_dir}/${part.value}")
```
nh-exa, line 147: this should be updated to use the same bootstrap asset mechanism as the other will use now

nh-exa (changes requested):

assets/infrastructure/stackit/cloudinit.tf
migroble: Whoops, forgot to remove that.

assets/infrastructure/stackit/cloudinit.tf
migroble: I've been trying to get this to work with the other mechanism and it turns out it's surprisingly non-trivial due to STACKIT's authentication model and how AWS's provider consumes the access and secret keys.

Basically, the access and secret keys for the S3 bucket are created at "apply-time", but the AWS provider expects them to be known "init-time", so we can't use the AWS provider as-is.

The alternatives we have are:
- Split the "apply" into two phases: "bootstrap bucket + credentials" and "everything else". This would require some refactoring of the launcher to support multi-phase deployments.
- Have the user create the credentials and pass them in via env vars. This implies more manual steps for the user (creating a credentials group and the credentials) on top of the already required ones (creating a project and the service account).
- Replace "aws_s3_object" with a script that uploads the files. This would work, but it's pretty ugly.
- Create our own Terraform provider with a "aws_s3_object"-like resource. Higher-effort and nicer version of the previous option.

I don't like any of these, to be honest. What do you think?

migroble: Looks like there's a minio (a generic S3-compatible server) provider that supports this use case and actually works. https://registry.terraform.io/providers/aminueza/minio

The provider is not "official", but seems pretty well supported.

tests/tests/deployment/test_standard_deployment.py
```diff
@@ -560,7 +566,7 @@ def test_remote_archive_registered(
     Then we should receive a successful response
     And the response should contain the 'default_archive' backup option
     """
-    if infra not in {"aws", "azure"}:
+    if infra not in {"aws", "azure", "stackit"}:
         pytest.skip("Remote archive verification is only supported for AWS/Azure today")
```
nh-exa, line 570: I think the exception here was for Exoscale and even that isn't true anymore.
So I hope this test doesn't need to skipped for any providers anymore. Got to find out

migroble: I could remove this check, but since I can't test on Exoscale at the moment, it might be best to make that change in a different PR.

assets/infrastructure/stackit/files/opt/exasol_launcher/scripts/stackit_registerS3ArchiveVolume.sh
```diff
@@ -0,0 +1,66 @@
...
+fi
+
+# Get the owner tuple for the database
+OWNER_TUPLE=$(confd_client db_info db_name: "$DB_NAME" --json | jq -c ".config.owner")
+if [ -z "$OWNER_TUPLE" ]; then
+  exit 1
+fi
+
+# Register the S3 bucket as a remote archive volume
+if ! confd_client remote_volume_add vol_type: s3 url: "$S3_URL" owner: "$OWNER_TUPLE" remote_volume_name: "$VOLUME_NAME"; then
```
nh-exa: just to be sure: no authentication/credentials needed here? is access managed by some equivalent of AWS instance roles?

Taskfile.yml
nh-exa, line 204: don't we need to pass the stack project id here too?

tests/tests/conftest.py
```diff
@@ -32,3 +40,8 @@ def exasol_path(request: pytest.FixtureRequest) -> str:
 @pytest.fixture(scope="session")
 def infra(request: pytest.FixtureRequest) -> str:
     return str(request.config.getoption("--infra"))
+
+
+@pytest.fixture(scope="session")
+def stackit_project_id(request: pytest.FixtureRequest) -> str:
+    return str(request.config.getoption("--stackit-project-id"))
```
nh-exa: I think any test or fixture using this fixture (like "custom_deployment" or "reusable_deployment") will now always receive a "str" that is either "None" (if the parameter doesn't exist) or the stackit project id.

So i think this method should return the type `None` if the option doesn't exist and not the string "None".

tests/tests/deployment/test_custom.py
```diff
@@ -85,6 +90,12 @@ def custom_deployment(
             marks=pytest.mark.provider_azure,
             id="azure-Standard_B2s",
         ),
+        pytest.param(
+            "stackit",
+            "g2i.1",
+            marks=pytest.mark.provider_azure,
```
nh-exa: `marks=pytest.mark.provider_stackit,`

nh-exa (changes requested):
Some issues.

Also, the README in the repo root the mentions all supported providers should now also mention STACKIT

assets/infrastructure/stackit/files/opt/exasol_launcher/scripts/stackit_registerS3ArchiveVolume.sh
migroble: I forgot to pass the access and secret keys. I think this is a gap our tests.

Taskfile.yml
migroble: Done.

tests/tests/deployment/test_custom.py
migroble: Done.

tests/tests/conftest.py
migroble: Done.

migroble:
> Also, the README in the repo root the mentions all supported providers should now also mention STACKIT

Done (in the docs PR #125).

nh-exa (approved):

PR #111, opened 2026-05-18
fix: exoscale cleanup tool reliability improvements

Gregory-Bull-Exasol:
Previously, if the compute was manually deleted from an exoscale deployment, the remaining resources could not be found.

Now, the resources will be found if any of ( compute | vpc | block-storage ) remain.

If we want to support finding the remaining resources, we will have to rely on regular expression's matching `deployment_id` to their names. I have not implemented this as it could cause problems if we change our deployment naming scheme, and I believe the current solution is sufficient.

### DIFF
--- a/tools/cleanup/internal/exoscale/collect.go
+++ b/tools/cleanup/internal/exoscale/collect.go
@@ -125,6 +125,7 @@ func CollectDeploymentDetails(
 			
 			nameStr := fullInst.Name
 			labels := fullInst.Labels
+			slog.Info("discovered instance", "id", instID, "name", nameStr, "labels", labels, "deploymentID", deploymentID)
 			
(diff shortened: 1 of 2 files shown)


nh-exa:
The errors i reported earlier were not about orphaned resources initially though. The compute instance were still there and running the clean up tool tried to delete e.g. the private network or the block storage of the instance  before the instance was terminated. So there is an ordering problem.
I had this with AWS before to. The order of termination is important and some of it was timing sensitive. E.g. you need to wait a few seconds after you terminated the instance so that AWS would have time internally update its state to allow e.g. the VPC to be removed. This stuff wasn't always synchronous.

nh-exa (approved):

PR #128, opened 2026-06-01
Support safe reuse of deployment directories

nh-exa:
## Summary

This change makes the default deployment directory safe to reuse without silently deploying stale preset state.

Users can now:
- rerun `install` with the same presets to retry a failed deployment
- update same-preset configuration on a freshly-initialized deployment (before deploy or after a clean destroy)
- inspect and reset configuration through `config get` and `config reset`
- switch presets only after explicit cleanup with `destroy --remove`
- recover abandoned local state with `remove` when cloud resources are already gone

### Always-on visibility of the active deployment directory

Every command logs the resolved deployment directory together with how it was resolved (`explicit`, `current`, or `default`), not just when the default location is used. This makes log trails and support tickets unambiguous about where the launcher is operating, even when the user did not pass `--deployment-dir`.

Example log entry (JSON formatter):

```json
{"level":"INFO","msg":"using deployment directory","path":"/Users/me/.exasol/personal/deployments/default","source":"default"}
```

### Confirmation prompts spell out the directory

`exasol remove` and `exasol destroy --remove` now name the exact local deployment directory before asking for confirmation, so the user can sanity-check the path before approving a destructive operation.

```
WARNING: This removes the local deployment directory. It does not destroy
cloud resources. Continue only if the resources were already deleted manually
or you accept that launcher-based cleanup may no longer be possible.

Local deployment directory to remove:
/Users/me/.exasol/personal/deployments/default

Proceed with local removal? [y/N]
```

```
WARNING: Destroying a deployment is an irreversible operation, and removes all
cloud resources - including all data.

Local deployment directory to remove after destroy:
/Users/me/.exasol/personal/deployments/default

Proceed with destroy? [y/N]
```

Plain `exasol destroy` (without `--remove`) continues to omit the local-removal section.

### Safe local removal with actionable errors

`exasol remove` (and the local-removal step of `destroy --remove`) now fails fast — before deleting anything — when the current execution context would make the deletion unsafe or partially-failing, especially on Windows:

- the current working directory is the deployment directory or a path inside it
- the running launcher binary lives inside the deployment directory

Errors name both the offending path and the deployment directory, and tell the user what to do:

```
deployment directory cannot be removed safely: current directory
"/Users/me/.exasol/personal/deployments/default/subdir" is inside deployment
directory "/Users/me/.exasol/personal/deployments/default"; change to another
directory and rerun
```

```
deployment directory cannot be removed safely: launcher binary
"/Users/me/.exasol/personal/deployments/default/bin/exasol" is inside
deployment directory "/Users/me/.exasol/personal/deployments/default";
move the launcher binary and rerun
```

Other deletion failures (permission errors, OS locks, etc.) now include the exact path that failed, instead of a bare file name:

```
failed to remove local deployment path
"/Users/me/.exasol/personal/deployments/default/infrastructure":
permission denied
```

## UX Examples

```bash
> exasol init aws --cluster-size 2
...
> exasol init aws --cluster-size 3
...
Active configuration:
  cluster-size = 3
...
configuration updated locally; run `exasol deploy` to apply these changes
> exasol deploy
```
Updates local configuration for a freshly-initialized AWS deployment, then applies it with deploy.

```bash
> exasol config get
Active configuration:
  cluster-size = 1
  data-volume-size = 100
  ...
> exasol config set --cluster-size 3
...
Active configuration:
  cluster-size = 3
...
configuration updated locally; run `exasol deploy` to apply these changes
> exasol deploy
```

Shows active configuration and updates selected values locally; `config set` and config reset` are only allowed in the initialized workflow state

```bash
> exasol install aws --cluster-size 2
...
> exasol destroy
...
> exasol install aws --cluster-size 3
```

Once a deployment may have cloud resources — running, stopped, deployment-failed, or interrupted during `deploy` or `destroy` — `config set` / `config reset` refuse with `exasol destroy` (or `exasol remove` if the cloud resources are confirmed gone) guidance, after which the user can change same-preset configuration and deploy again.

```bash
> exasol install aws
> exasol install stackit
Error: deployment directory is initialized with different presets: existing infrastructure "aws", installation "ubuntu"; requested infrastructure "stackit", installation "ubuntu".
Run `exasol destroy --remove` before initializing different presets, or run `exasol remove` if the cloud resources are already gone
```

Refuses the second command because the deployment directory is initialized with different presets.

```bash
> exasol destroy --remove
2026-06-03 16:13:36 INF using deployment directory path=/home/nh/.exasol/personal/deployments/default source=default
WARNING: Destroying a deployment is an irreversible operation, and removes all cloud resources - including all data.

Local deployment directory to remove after destroy:
/home/nh/.exasol/personal/deployments/default

Proceed with destroy? [y/N]: y
...
> exasol init stackit
```

Destroys cloud resources, removes the local deployment directory, then allows a different preset.

```bash
> exasol remove 
2026-06-03 16:17:54 INF using deployment directory path=/home/nh/.exasol/personal/deployments/default source=default
WARNING: This removes the local deployment directory. It does not destroy cloud resources. Continue only if the resources were already deleted manually or you accept that launcher-based cleanup may no longer be possible.

Local deployment directory to remove:
/home/nh/.exasol/personal/deployments/default

Proceed with local removal? [y/N]: y
```

Removes abandoned local deployment state without attempting cloud cleanup.

## Verification

- `openspec validate support-deployment-reconfiguration --strict`
- `task all`


cmd/exasol/infra_variables.go
```diff
@@ -262,3 +310,55 @@ func scanInfrastructurePresetSelection(args []string) (*deploy.PresetRef, error)
...
+
+	defaultDir, err := defaultDeploymentDir()
+	if err != nil {
+		return config.DeploymentDir{}, err
+	}
+
+	return config.NewDeploymentDir(filepath.Clean(defaultDir)), nil
+}
+
+func deploymentDirFromRawArgsIfAvailable(args []string) (config.DeploymentDir, bool) {
```
migroble: I think this function can be removed, all its callers already check for failure.

cmd/exasol/infra_variables.go
```diff
@@ -127,30 +129,55 @@ func prepareInfrastructureVariableFlags(args []string) error {
...
+			}
+		}
 	}
-	if preset == nil {
+
+	return prepareConfigSetInfrastructureVariableFlags(args)
+}
+
+func prepareConfigSetInfrastructureVariableFlags(args []string) error {
+	if !rawCommandSequenceSelected(args, "config", "set") {
```
migroble: This feels pretty hacky. Does cobra provide functionality to parse the cmdline flags before executing? That would allow us to refactor this entire file to be a bit nicer since we wouldn't need to "guess" which flag is what or have command-specific workarounds.

cmd/exasol/infra_variables.go
```diff
@@ -262,3 +310,55 @@ func scanInfrastructurePresetSelection(args []string) (*deploy.PresetRef, error)
...
+
+	return false
+}
+
+func deploymentDirFromRawArgs(args []string) (config.DeploymentDir, error) {
+	for i, tok := range args {
+		if tok == "--"+deploymentDirFlagName && i+1 < len(args) {
+			return config.NewDeploymentDir(args[i+1]), nil
+		}
+		if value, ok := strings.CutPrefix(tok, "--"+deploymentDirFlagName+"="); ok {
```
migroble: The `--deployment-dir` flag can be specified without the "=" as well.

cmd/exasol/install_variables.go
```diff
@@ -62,24 +63,51 @@ func prepareInstallationVariableFlags(args []string) error {
...
+			}
+		}
 	}
-	if preset == nil {
+
+	return prepareConfigSetInstallationVariableFlags(args)
+}
+
+func prepareConfigSetInstallationVariableFlags(args []string) error {
+	if !rawCommandSequenceSelected(args, "config", "set") {
```
migroble: Same as in `infra_variables.go`, this feels hacky.

cmd/exasol/root.go
```diff
@@ -82,13 +83,10 @@ var rootCmd = &cobra.Command{
 			}
 		}
 
-		// Best-effort version update hint (non-blocking; logs only when an update is available).
+		// Best-effort version update hint (non-blocking; terminal-only when available).
```
migroble, line 86: This seems like a search and replace gone wrong.

internal/deploy/destroy.go
```diff
@@ -89,5 +94,228 @@ func Destroy(ctx context.Context, deployment config.DeploymentDir, verbose bool)
...
+	}
+
+	entries, err := os.ReadDir(deployment.Root())
+	if err != nil {
+		return err
+	}
+
+	for _, entry := range entries {
+		name := entry.Name()
+		if strings.HasPrefix(name, ".dirmutex.") {
```
migroble: I'm not a big fan of hardcoding this prefix here, can we instead export the constant or a `IsDirectoryMutexMarker(str)` from `directorymutex`?

tests/tests/integration/test_deployment_directory_resolution.py
```diff
@@ -91,6 +95,29 @@ def test_status_reports_uninitialized_explicit_deployment_dir(
...
     assert data["deploymentDir"] == str(deployment_dir)
+    _assert_deployment_dir_logged(result.stderr, deployment_dir, "explicit")
+
+
+def test_status_logs_current_deployment_dir(exasol_path: str, tmp_path: Path) -> None:
+    # Given the current working directory is a recognized deployment directory
+    deployment_dir = tmp_path / "deployment"
+    deployment_dir.mkdir()
+    (deployment_dir / ".exasolLauncher.version").write_text("0.0.0")
+    launcher = str(Path(exasol_path).resolve())
```
migroble, line 105: I think it might be more robust to initialize the deployment directory with the launcher instead.

tests/tests/integration/test_reconfiguration.py
```diff
@@ -0,0 +1,811 @@
...
+    for preset in presets[:count]:
+        preset_id = preset.get("id")
+        if not isinstance(preset_id, str):
+            pytest.skip("infrastructure preset list contains invalid IDs")
+        ids.append(preset_id)
+
+    return ids
+
+
+def _write_configurable_no_resource_presets(tmp_path: Path) -> tuple[Path, Path]:
```
migroble: Same here and a few tests in this file, initializing with the launcher should be less brittle.

tests/tests/integration/test_reconfiguration.py
```diff
@@ -0,0 +1,811 @@
...
+    )
+    deployment_log = (deployment_dir / "deployment.log").read_text()
+    assert "configuration updated locally" not in deployment_log
+    assert "run `exasol deploy` to apply these changes" not in deployment_log
+    assert state_file.read_text() == "partial state"
+    tfvars = (deployment_dir / "infrastructure" / "vars.tfvars").read_text()
+    assert "cluster_size" in tfvars
+    assert "= 3" in tfvars
+    assert "instance_type" in tfvars
+    assert '"custom-instance"' in tfvars
```
migroble: I don't like asserting the entire output text, I think we should only assert that the new value is printed, even checking the tfvars file feels too white-boxy for my taste.

Same feedback applies to most of these tests.

migroble (changes requested):
I really like the UX changes. The terminal notices at the end are super nice.

Some minor issues with the implementation and gripes with the tests.

cmd/exasol/infra_variables.go
nh-exa: done

cmd/exasol/infra_variables.go
nh-exa: i tried to improve it (and i did) and stumbled on some cobra crash while doing so. it's better now though

cmd/exasol/infra_variables.go
nh-exa: ok

cmd/exasol/install_variables.go
nh-exa: done

cmd/exasol/root.go
nh-exa: yes

internal/deploy/destroy.go
nh-exa: agreed

tests/tests/integration/test_deployment_directory_resolution.py
nh-exa: yeah, more code but also more SSOT

tests/tests/integration/test_reconfiguration.py
nh-exa: done

tests/tests/integration/test_reconfiguration.py
nh-exa: LLMs are over-eager when writing tests. I toned it down a bit.

cmd/exasol/root.go
nh-exa: yes

cmd/exasol/config.go
```diff
@@ -0,0 +1,212 @@
...
+var configResetAll bool
+
+var configCmd = &cobra.Command{
+	Use:     "config",
+	Short:   configCmdShortDesc,
+	Long:    configCmdLongDesc,
+	GroupID: rootCmdGroupEssential,
+}
+
+var configGetCmd = &cobra.Command{
```
wizenink, line 36: We should probably annotate all of the new commands. Otherwise, compatibility gates and deployment log annotations are not going to be present for them

internal/deploy/configure.go
```diff
@@ -0,0 +1,126 @@
...
+	if clusterIdentity == "" {
+		return errors.New("deployment state is missing cluster identity")
+	}
+
+	artifactDir := deployment.RelativeInfrastructureArtifactDir()
+	infrastructureValues["infrastructure_artifact_dir"] = artifactDir
+	infrastructureValues["installation_preset_dir"] = deployment.RelativeInstallationPresetDir()
+	infrastructureValues["deployment_id"] = deploymentId
+	infrastructureValues["cluster_identity"] = clusterIdentity
+	infrastructureValues["deployment_created_at"] = time.Now().UTC().Format(time.RFC3339)
```
wizenink, line 87: We should store `deployment_created_at` and not rewrite it, it is not really a new deployment as such, and can be misleading for users looking at resources in the cloud

wizenink (changes requested):

internal/deploy/configure.go
nh-exa: this actually hinted me at a deeper design issues around these variables which i have to address to fix this issue now

cmd/exasol/config.go
nh-exa: doing that

internal/deploy/configure.go
nh-exa: While there is a lot to improve around handling of the preset values - I've decided to this fix this in a separate refactor.
The issue you brought up here is correct but not critical atm because you can't `configure` a deployment unless it isn't live. We might eventually allow changing configuration for live deployment if we figure out a mechanism to distinguish configuration which actually can be changed on live deployments from config which can't.
there might just be different kinds of configuration.
And the deployment_created_at value you see is strictly only required for the opentofu backend. there is an architecture layer violation here I'd like to fix too. But not in this PR.
I'll have a followup next week

migroble (approved):

wizenink (approved):

PR #129, opened 2026-06-02
Implement local infrastructure preset and backend with the exasol-local-vm-runner

wizenink:
## Description

Adds a local Exasol DB deployment backend managed by the launcher. The launcher now supports deploying, starting, stopping, destroying, connecting to, and opening shells for local VM deployments using the macOS Apple Silicon runner. The runner is embedded onto the binary at build time.
This PR also updates presets to expose AdminUI metadata and updates AdminUI metadata handling so cloud deployments can expose it at will while local does not include any web UI.

## Related Issue

Fixes #

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Added built-in `local` infrastructure and `local` installation presets for local Exasol DB VM deployments.
- Added launcher-managed local runtime lifecycle: deploy, start, stop, destroy, SSH key/share setup, runner state parsing, and local artifact generation.
- Added build-time runner embedding/staging via `assets/localruntimebin` and Taskfile support.
- Added local shell support for VM host shell and DB container shell.
- Exposed local CPU, memory, and DB shared-memory sizing through the runner start contract.
- Updated local `deployment.json` to use endpoint connection metadata without cloud-style `nodes`.
- Added optional Admin UI deployment metadata for cloud presets, while omitting Admin UI metadata from local deployments.

## Testing

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

## Additional Notes

Changes left for a follow-up PR:

* simplify download-stage steps
* use resources.yml mechanism to download the vm
* simplify SSH key generation and authorized_keys
* ssh commands and multiline strings as embedded assets instead of hardcoding as strings


.github/actions/build/action.yml
```diff
@@ -1,13 +1,17 @@
 name: build
 description: Build the project
+inputs:
+  local-runner-token:
+    description: GitHub token with read access to exasol-nano-vm releases
```
migroble: `exasol-nano-vm` should be updated to `exasol-local-vm`.

nh-exa: @wizenink Didn't you want to grep for that?

Also, we don't need that token anymore, do we?

internal/deploy/local_backend.go
```diff
@@ -0,0 +1,436 @@
...
+		"rootfs=$(podman mount \"$container_name\") || exit $?;",
+		"pid=$(podman inspect \"$container_name\" --format '{{.State.Pid}}') || exit $?;",
+		"echo \"Exasol Local database container image does not include a shell; " +
+			"using VM shell with container rootfs mounted as working directory.\";",
+		"echo \"Container rootfs: $rootfs\";",
+		"cd \"$rootfs\" || exit $?;",
+		"if [ -n \"$pid\" ] && [ \"$pid\" != \"0\" ]; then",
+		"exec nsenter --target \"$pid\" --uts --ipc --net /bin/sh;",
+		"fi;",
+		"exec /bin/sh",
```
migroble, line 298: Could this be made more legible with a multiline string + `fmt.Sprintf`  instead?

internal/deploy/local_backend.go
```diff
@@ -0,0 +1,436 @@
...
+		"container_name=" + localDBContainerName + ";",
+		"if ! podman container exists \"$container_name\"; then",
+		"container_name=" + localLegacyDBContainerName + ";",
+		"fi;",
+		"if ! podman container exists \"$container_name\"; then",
+		"echo \"Exasol Local database container not found\" >&2;",
+		"podman ps -a >&2;",
+		"exit 125;",
+		"fi;",
+		"for shell_path in /bin/bash /usr/bin/bash /bin/sh /usr/bin/sh /bin/ash /usr/bin/ash; do",
```
migroble, line 284: We have control over the contents of the DB container, I don't think we need to check which shell we are using, do we? If so, we also don't need the fallback shell below.

internal/deploy/local_backend.go
```diff
@@ -0,0 +1,436 @@
...
+		"echo \"Container rootfs: $rootfs\";",
+		"cd \"$rootfs\" || exit $?;",
+		"if [ -n \"$pid\" ] && [ \"$pid\" != \"0\" ]; then",
+		"exec nsenter --target \"$pid\" --uts --ipc --net /bin/sh;",
+		"fi;",
+		"exec /bin/sh",
+	}, " ")
+}
+
+func localSSHRemoteUnsafe(deployment config.DeploymentDir) (*remote.SSHRemote, error) {
```
migroble, line 302: What is unsafe about this function?

internal/deploy/ready.go
```diff
@@ -80,8 +80,8 @@ func WaitForDatabaseStarted(
 		ctx,
 		deployment,
 		WaitParams{
-			InitialBackoff: StartedInitialBackoff,
-			MaxBackoff:     StartedMaxBackoff,
+			InitialBackoff: DatabaseStartedInitialBackoff,
+			MaxBackoff:     DatabaseStartedMaxBackoff,
```
loaj84: WaitForDatabaseStarted is also used by the cloud path (tofu_backend.go), so this drops cloud's polling interval from 10s/60s to 2s/8s too. Is that intended for cloud as well?

internal/deploy/local_backend.go
```diff
@@ -0,0 +1,436 @@
...
+) error {
+	sshRemote, err := localSSHRemoteUnsafe(b.deployment)
+	if err != nil {
+		return err
+	}
+
+	return sshRemote.Shell(ctx, os.Stdout, os.Stderr)
+}
+
+func (b *localBackend) OpenCOSShell(ctx context.Context) error {
```
nh-exa, line 262: Reminder for myself: This method really needs a new name

internal/deploy/local_backend.go
```diff
@@ -0,0 +1,436 @@
...
+	}
+
+	command := localContainerShellCommand()
+
+	return sshRemote.RunInteractiveCommand(ctx, command, os.Stdout, os.Stderr)
+}
+
+func localContainerShellCommand() string {
+	return strings.Join([]string{
+		"container_name=" + localDBContainerName + ";",
```
nh-exa, line 275: i feel like this code should be either part of the VM or at least an embedded Go asset. What do you think?

internal/localruntime/runtime.go
```diff
@@ -0,0 +1,585 @@
...
+) (*State, error) {
+	runtime := newRuntime(deployment, runtimeConfig)
+	if err := runtime.prepareForDeploy(ctx, out, outErr); err != nil {
+		return nil, err
+	}
+
+	return runtime.start(ctx, out, outErr)
+}
+
+func Start(
```
migroble, line 130: These methods are part of the `deploymentBackend` interface, they should probably be in `local_backend.go`.

internal/localruntime/runtime.go
```diff
@@ -0,0 +1,585 @@
...
+}
+
+func (runtime *localRuntime) ensureSSHKey() error {
+	if _, err := os.Stat(runtime.paths.PrivateKeyPath); err == nil {
+		return runtime.writeAuthorizedKeyFromPrivateKey()
+	} else if !errors.Is(err, os.ErrNotExist) {
+		return fmt.Errorf("failed to inspect local SSH key: %w", err)
+	}
+
+	privateKey, err := rsa.GenerateKey(rand.Reader, sshKeyBits)
```
loaj84, line 392: Any reason for RSA-4096 here over ed25519? For a freshly generated key, ed25519 is smaller and faster.

internal/localruntime/runtime.go
```diff
@@ -0,0 +1,585 @@
...
+		VMIP:                   state.VMIP,
+		SSHPort:                state.Ports.SSH,
+		DBPort:                 state.Ports.DB,
+		UIPort:                 state.Ports.UI,
+		PrivateKeyPath:         runtime.paths.PrivateKeyPath,
+		PrivateKeyRelativePath: keyFile,
+	}, nil
+}
+
+func (runtime *localRuntime) ensureRunnerExecutable() error {
```
migroble, line 262: This entire area of the code can be massively simplified if `writeEmbeddedRunner` only writes the binary if it doesn't already exist. Additionally, why do we have an environment variable for the runner path? We are embedding it into the launcher and then putting it into the deployment directory, right?

tools/localrunner/main.go
```diff
@@ -0,0 +1,624 @@
...
+	if err := downloadRunner(ctx, config, targetPath); err != nil {
+		return "", err
+	}
+
+	return targetPath, nil
+}
+
+func fallbackRunnerPaths() []string {
+	return []string{
+		"exasol-local-vm/ci-downloads/release/launcher",
```
nh-exa: what are these fallback paths?

internal/localruntime/runtime.go
```diff
@@ -0,0 +1,585 @@
...
+	return runtime.prepare(ctx, runtime.ensureRunnerExecutable, out, outErr)
+}
+
+func (runtime *localRuntime) prepareForStart(
+	ctx context.Context,
+	out, outErr io.Writer,
+) error {
+	return runtime.prepare(ctx, runtime.ensureRunnerExecutableIfMissing, out, outErr)
+}
+
```
migroble: These two helpers can be removed if you change how the runner executable is handled.

tools/localrunner/main.go
nh-exa, line 1: i think i want a readme explaining what is happening here. where things are downloaded from, what "stage" is etc.

internal/localruntime/runtime.go
```diff
@@ -0,0 +1,585 @@
...
+	if err := os.MkdirAll(runtime.paths.ShareDir, dirMode); err != nil {
+		return fmt.Errorf("failed to create local managed share: %w", err)
+	}
+
+	authorizedKeyPath := filepath.Join(runtime.paths.ShareDir, authorizedKeysFile)
+
+	return appendPublicKeyFromPrivateKey(runtime.paths.PrivateKeyPath, authorizedKeyPath)
+}
+
+func appendPublicKeyFromPrivateKey(privateKeyPath, authorizedKeyPath string) error {
```
migroble, line 424: This code could be simplified a lot if instead of appending the authorized key we overwrite the authorized keys. I don't see any reason why we would ever have more than one key to access the VM anyway.

internal/deploy/local_backend.go
nh-exa: I proposed an embedded asset and parameters even. but yes, using fmt.Sprintf seems to be canonical choice. 
Golang never added string interpolation unfortunately and they cannot add it now since it would break old code.

internal/deploy/local_backend.go
nh-exa: Well, "we" as Exasol do. "We" as the team do not. So I don't mind this.

tools/localrunner/main.go
```diff
@@ -0,0 +1,624 @@
...
+func run(ctx context.Context, args []string) error {
+	if len(args) == 0 {
+		return errors.New("expected subcommand: placeholder, download, or stage")
+	}
+
+	switch args[0] {
+	case "placeholder":
+		return runPlaceholder(args[1:])
+	case "download":
+		return runDownload(ctx, args[1:])
```
migroble, line 67: You could use the new `resources.yaml` mechanism to download the VM, it implements the checksum validation and all that, so this file would end up a whole lot smaller.

Taskfile.yml
```diff
@@ -23,10 +23,51 @@ tasks:
...
+      LOCAL_RUNNER_ASSET: '{{default "mac-runner-aarch64.zip" .LOCAL_RUNNER_ASSET}}'
+    cmds:
+      - >-
+        go run ./tools/localrunner stage
+        -repo {{.LOCAL_RUNNER_REPO}}
+        -version {{.LOCAL_RUNNER_VERSION}}
+        -asset {{.LOCAL_RUNNER_ASSET}}
+        -runner-path "{{.RUNNER_PATH}}"
+        -target assets/localruntimebin/generated/darwin/arm64/mac-runner-aarch64
+        -download-dir assets/localruntimebin/downloads/darwin/arm64
```
migroble, line 48: I don't think this "stage" step is necessary, we could move the runner into the right path after downloading it.

internal/localruntime/runtime.go
wizenink: The VM runner connects to the VM and sends `poweroff` over ssh, so we would have to either use the same key from the runner, or get our key into the runner aswell

internal/deploy/local_backend.go
wizenink: Bad grepping led me here, I got .git ignored from rg usually, wil correct that

internal/deploy/local_backend.go
migroble: That's a good point, but I think it's OK to assume we will have sh or whatever shell we are using in the DB container since it would simplify this a lot.

internal/deploy/local_backend.go
migroble: An embedded asset seems like an even better choice, that way we could lint it and it's more discoverable -- I really don't like hardcoding code in one language inside other languages.

internal/deploy/local_backend.go
wizenink: It assumes that the caller has already acquired the lock over the deployment data. It follows the same convention as other "unsafe" functions. However, I agree that the name is kinda misleading

migroble: I think we should document that assumption, a function with `Unsafe` in the name but no safe usage requirements is scary.

internal/deploy/ready.go
wizenink: Good catch. I'd change that to use 2/8 for local and the usual 10/60 for cloud. Not pretty, but we want local to start as quickly as possible

internal/deploy/local_backend.go
wizenink: well, actually the DB container itself does not have any shell, it's extremely barebones

internal/localruntime/runtime.go
wizenink: Totally right, will change it to ed25519

internal/localruntime/runtime.go
migroble: I think the VM runner should take the SSH key by parameter then, that way we only have one instance of "code to generate keys" and "code to set the authorized keys" instead of one per repo and we can avoid some unnecessary complexity.

internal/localruntime/runtime.go
wizenink: Good point on the env var. Runtime EXASOL_LOCAL_RUNNER_PATH is a leftover from before the runner was embedded.
As for the second part, I originally wrote deploy to update the runner and start/stop to not do it so a user could update to a new version on a deployment, but `deploy` cannot be called into an undestroyed deployment dir anyways, so yes, that'd be a nice simplification

tools/localrunner/main.go
wizenink: That was a leftover from development, I'll remove it

nh-exa (dismissed):

nh-exa:
Alright, this can go in.
Let's not forget about the follow up work though.

nh-exa (approved):

migroble (approved):

PR #130, opened 2026-06-03
fix: Fix bug where results larger 1,000 rows were not fetched

juergen-albertsen-exasol:
## Description

`exasol connect` failed on any query returning more than the Exasol websocket inline limit (~1,000 rows). For example, `SELECT * FROM PRODUCTS` on the sample data errored with:

ERR the result set doesn't have expected number of columns: expected=1000000 got=0

The launcher executed queries via the driver's raw `SimpleExec` and parsed only the inline response, never fetching the result-set handle the server returns for larger results. A secondary bug made the message nonsensical — it printed the row count where the column count belonged.

This PR fixes the retrieval and adds a row-display cap so a large `SELECT` doesn't flood the terminal, while keeping non-interactive (piped) output complete.

## Related Issue

N/A

## Type of Change

- [x] `fix`: Bug fix
- [x] `feat`: New feature (`--max-rows`)

## Changes Made

- Execute queries through the driver's `QueryContext`/`driver.Rows`, which transparently fetches result-set chunks and closes the server-side handle — fixing large result sets.
- Add a row cap: interactive (TTY) sessions default to 100 rows with a truncation note on **stderr**; piped/non-interactive sessions return the full result set. New `--max-rows N` flag overrides either mode (`0` = unlimited).
- Stop fetching once the cap is reached, so unshown rows are never transferred.
- Remove the obsolete manual response parsing and the swapped-variable error message; `QueryResult` is now a plain data holder with a `Truncated()` flag.
- Integer (scale-0 `DECIMAL`) values now render as `1000000` instead of the float artifact `1e+06`.
- Document `--max-rows` and the interactive preview cap in the README; spec captured under `openspec/changes/fix-connect-large-result-sets`.

## Testing

### Test Details

Unit tests cover: unlimited fetch returning all rows, truncation stopping at the cap (with a single peek to detect "more available"), result-set handle closed exactly once, integer rendering, non-resultset statements (e.g. `OPEN SCHEMA`), query errors, limit resolution (interactive → 100, non-interactive → unlimited, explicit `--max-rows`), and that the footer goes to stderr only when truncated. `go build ./...`, full `go test ./...`, and `golangci-lint` on the touched packages all pass.

Manual verification against a live deployment done:

  ```bash
  # interactive: shows 100 rows + a stderr footer
  exasol connect
  > OPEN SCHEMA sample;
  > SELECT * FROM products;

  # return everything, even interactively
  exasol connect --max-rows 0

  # piped input returns the full result set (no cap)
  echo "SELECT * FROM PRODUCTS;" | exasol connect

  # explicit cap
  echo "SELECT * FROM PRODUCTS;" | exasol connect --max-rows 1000
  ```

## Checklist

  - [x] Code follows the project's [coding guidelines](https://github.com/exasol/exasol-personal/compare/doc/best_practices.md)
  - [x] Ran task fmt and task lint
  - [x] Added/updated tests
  - [x] All tests pass locally

##  Additional Notes

  The truncation footer reports the number of rows shown but not an exact total — the driver's driver.Rows doesn't expose the result-set size (NumRows is unexported with no accessor), and obtaining it would require either a second execution or reimplementing the fetch protocol. Surfacing an exact total is noted as a
  possible follow-up needing an upstream driver change.


nh-exa (approved):
Good contribution.

PR #133, opened 2026-06-03
feat: Download OpenTofu at runtime

migroble:
## Description

Currently OpenTofu is embedded into the binary and extracted at runtime. This bloats the binary and makes it awkward to build. This PR implements a "resource manager" that takes in a specification of a set of resources and lazily fetches them when they are requested.

The schema of the resources specification allows us to specify whether the artifact should be uncompressed and can select an artifact URL based on OS and architecture or a (currently unused) `common` variant which applies to all OSs and architectures.

## Related Issue

[SPOT-27111](https://exasol.atlassian.net/browse/SPOT-27111)

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Launcher:
  - Added a "resource manager" which handles downloading, validating (sha256 checksum), extracting (currently only `.tar.gz`), and caching (per-deployment, not globally) resources the launcher needs (currently only OpenTofu).
  - Added the resources specification file.
  - Modified the `tofu.Config.TofuBinaryPath()` method to request the `tofu` resource.
  - Removed compile-time OpenTofu downloading and embedding.
- Tools:
  - Removed `downloadtofu` tool.
  - Added `tofu` tool to download and call OpenTofu (used for `task fmt-terraform`).
  - Modified `lockfileupdate` tool to download OpenTofu at runtime.

## Testing

### Test Details

I manually executed the `task tofu-lockfile-update` and `task lint-terraform` tasks, which internally use OpenTofu, and checked that they still work as expected. The rest was tested by running our tests.

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

## Additional Notes

The resources specification file is embedded in the binary, but we might want to allow users to override it with an external file passed in via argument.

We might also want to cache resources per-user rather than per-deployment.

[SPOT-27111]: https://exasol.atlassian.net/browse/SPOT-27111?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/resourcemgr/manager_test.go
```diff
@@ -0,0 +1,356 @@
+// Copyright 2026 Exasol AG
+// SPDX-License-Identifier: MIT
+
+package resourcemgr
```
nh-exa: I think I prefer if we didn't abbreviate things unnecessarily.
Also, "resource manager" as a name appears a little too abstract. First question I'd have is: what is a "resource" supposed to be? It feels like calling a cat "rodent lifetime manager" :D

I've seen (and implemented) like 15 resource managers in my life and none of those were like this.

assets/resources/resources.yaml
```diff
@@ -0,0 +1,18 @@
+tofu:
```
nh-exa, line 1: one of my more earlier visions was that the presets themselves would have a section in their manifest in which they declare which tools (resources) they need.

By embedding and essentially hard-coding this, we would not allow presets which are not embedded (and we support that already) to use resources which we don't declare here, right?

Our PM hasn't asked for that yet but that will have to happen eventually. Think about what happens once people want to maintain their own presets in their own repos.
So at the very least, there must be straight path towards supporting that, if we don't support it right now. And for that, I'd rather not completely change the architecture of this feature later.

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,122 @@
...
+
+## Purpose
+
+Define how Exasol Personal manages runtime resources in the deployment directory through an embedded resource spec, a deployment-local cache, and a generic resource manager API.
+
+## ADDED Requirements
+
+### Requirement: Launcher SHALL resolve runtime resources through a resource manager
+
+The launcher SHALL request runtime resources through a resource manager instead of writing compile-time embedded binaries into the deployment directory.
```
nh-exa: When working in OpenSpec or with LLMs in general I'd to keep implementation details out of the specs.
The fact that there is something called a "resource manager" is of no concern for the user. What matters for the user is that tool dependencies for the preset are downloaded, they this is obvious that it is happening and maybe that this can be configured so if a version of tofu is deprecated, another can be chosen.
that kind of stuff.

There really is little point in repeating implementation details here. 
Think about how our PM would write requirements. Forget that you are developer here for a second.

nh-exa (changes requested):

internal/resourcemgr/manager_test.go
migroble: I guess the AI got tired of typing "manager" and went with "mgr". :P

I thought about calling it "downloader" or something like that, but since it also manages the lifecycle of the downloaded artifacts, it felt too narrow. I'm open to ideas here.

assets/resources/resources.yaml
migroble: We could allow presets to provide a `resources.yaml` to override the "default" one. It would just be a matter of merging the maps before giving the spec to the manager, so that should be compatible with this architecture.

I do like the idea of having a default `resources.yaml` to avoid having to specify the `tofu` resource in (nearly) every single preset.

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
migroble: I completely missed that the very first requirement is basically "implement the thing the requirement requires"... I should definitely review the generated spec a lot closer, I mostly focused on the scenarios.

nh-exa: It's not just that though. The word "resource manager" should not appear in this document at all. That is an implementation detail. 
The requirements are about observable behavior and the launcher should be treated as a black box here. 
I am suspecting this happens when you go to your LLM with a design first and not requirements first.
There might be something like the resource manager in the implementation in the end, but you'd only see this in the design documents if at all - which you should discuss only after the requirements are clear.

migroble: That's a good point, I started with the use-case and some requirements then jumped straight to design to play around with the idea more concretely. I do prefer that "prototype-y" approach more since it gives me something concrete to get angry about and refactoring or scrapping ideas is cheap, but maybe it doesn't mesh well with the SDD workflow as much. Definitely something I should play around with more.

internal/resourcemgr/manager_test.go
migroble: The LLM's best suggestion was `runtimeartifacts`, which seems reasonable to me.

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,116 @@
...
+- **AND** it does not need to know the artifact URL, checksum, extraction behavior, or platform mapping
+
+#### Scenario: Cached resource path is reused
+- **WHEN** a runtime resource already exists in the local cache and matches the current definition for that resource
+- **THEN** the cached resource is reused
+- **AND** the artifact is not downloaded again
+
+### Requirement: Runtime resources SHALL be cached in the deployment directory
+
+The application SHALL store downloaded artifacts and cache metadata inside the deployment directory.
```
nh-exa, line 25: needs a followup PR to put the cache to `~/.exasol/` somewhere

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,116 @@
...
+- **THEN** the application downloads it
+- **AND** stores it in the deployment directory
+- **AND** updates the cache metadata
+
+#### Scenario: Resource artifacts are scoped by resource name
+- **WHEN** the application stores a resource artifact
+- **THEN** it stores the artifact in the deployment-local cache for that resource name
+- **AND** the artifact path is isolated from other resources
+
+### Requirement: Platform-specific artifacts SHALL be selected by current platform
```
nh-exa: runtime platform

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,116 @@
...
+- **WHEN** a resource defines a shared artifact
+- **THEN** that artifact is used for all supported platforms
+- **AND** the resource does not need platform-specific entries
+
+#### Scenario: Platform-specific artifact resource is valid
+- **WHEN** a resource defines one or more platform-specific artifact entries for the current platform
+- **THEN** the matching entry for the current platform is used
+- **AND** the resource does not define a shared artifact
+
+#### Scenario: Mixing common and platform-specific artifacts is rejected
```
nh-exa: lets not implement the common/shared resource/artfifact defintion until we know what we need exactly (and if at all)

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,116 @@
...
+- **WHEN** a resource defines both a shared artifact and platform-specific artifact entries
+- **THEN** parsing or validation fails
+- **AND** the error explains that a resource must use only one download selection mode
+
+### Requirement: Downloaded artifacts SHALL be verified before use
+
+The application SHALL verify the downloaded artifact using the expected checksum before it is used or extracted.
```
nh-exa: verification after download should be enough

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,116 @@
...
+- **AND** the error includes the actual value
+- **AND** the error identifies the resource being resolved
+
+#### Scenario: Valid checksum permits use of the cached artifact
+- **WHEN** the downloaded artifact checksum matches the expected value for the resource
+- **THEN** the artifact is accepted for caching or extraction
+
+### Requirement: Archive resources SHALL be optionally extracted
+
+The application SHALL support optional archive extraction when a resource requests it.
```
nh-exa: narrow this to "tar.gz"

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
```diff
@@ -0,0 +1,116 @@
+# runtime-resource-management Specification
```
nh-exa, line 1: We need to be able to specify the resource filename/path into the extracted archive to accommodate different filename for the different platforms (e.g. "tofu.exe" on Windows)

nh-exa (approved):
Good minimal scope implementation.
Waiting for the next PRs here :-)

openspec/changes/archive/2026-06-03-runtime-resource-management/specs/runtime-resource-management/spec.md
nh-exa: I thought about it and i think i'd prefer if we had this global cache sooner than later. I might test a lot of the deployment locally and it would annoy me greatly if I see it downloading the binary each time.

migroble: Working on that next.

PR #134, opened 2026-06-08
feat: Add per-user artifact cache

migroble:
## Description

Currently, resources are downloaded and cached per deployment directory. This PR adds a per-user artifact cache in the directory returned by `os.UserCacheDir()`.

The layout of the cache is:

```
.exasol
└── personal
    └── runtime-artifacts
        ├── downloads
        │   └── tmp.<hash>-<random>
        ├── artifacts
        │   └── <Resource ID>
        │       └── <OS>_<Arch>
        │           └── <hash>
        │               └── <Artifact>
        └── index.json
```

Where `downloads` contains files while they are being downloaded,  `index.json` contains information about the contents of the cache, and the `<hash>` is derived from:
- Logical runtime artifact ID.
- Platform.
- Source URL.
- Expected checksum.
- Extraction flag.
- Download path override.
- Resource path within an archive.

This PR also adds the following commands to manage the cache:

- `exasol cache list`: Lists the contents of the cache.
- `exasol cache clean --partial-downloads`: Removes artifacts which haven't finished downloading.
- `exasol cache clean --invalid`: Removes artifacts which have incorrect checksums.
- `exasol cache clean --all`: Wipes the cache.
- `exasol cache unlock`: Forcefully unlocks the cache (dangerous!).
- `exasol diag cache`: Outputs some information about the state of the cache.

All the `exasol cache clean` variants accept a `--dry-run` flag.

## Related Issue

[SPOT-31029](https://exasol.atlassian.net/browse/SPOT-31029)

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Added per-user resource cache.
- Removed per-deployment resource cache.
- Added `exasol cache <list,clean,unlock>` commands.
- Added `exasol diag cache` command.

## Testing

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

[SPOT-30669]: https://exasol.atlassian.net/browse/SPOT-30669?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

[SPOT-31029]: https://exasol.atlassian.net/browse/SPOT-31029?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/internal/deploy/locking.go
+++ b/internal/deploy/locking.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/exasol/exasol-personal/internal/config"
 	"github.com/exasol/exasol-personal/internal/directorymutex"
-	"github.com/exasol/exasol-personal/internal/util"
 )
 
 const lockUnavailableMessage = "Deployment directory is locked by another operation. " +
@@ -65,53 +64,29 @@ func withDeploymentLock(
 		return err
 	}
 
+	wrapped := func(any) error {
+		return function(deployment)
+	}
+
 	if exclusive {
-		err = mutex.AcquireExclusive(ctx)
+		err = mutex.WithExclusive(ctx, nil, wrapped)
 	} else {
-		err = mutex.AcquireShared(ctx)
+		err = mutex.WithShared(ctx, nil, wrapped)
 	}
 
 	if err != nil {
-		return mapLockAcquireError(ctx, deployment, err)
+		return mapDeploymentLockError(deployment, err)
 	}
 
-	releaseCtx := context.WithoutCancel(ctx)
-	release := func() error {
-		if exclusive {
-			return mutex.ReleaseExclusive(releaseCtx)
-		}
-
-		return mutex.ReleaseShared(releaseCtx)
-	}
-	releaseOnce := releaseOnInterruptOnce(release)
-
-	var callbackErr error
-	defer func() {
-		releaseErr := releaseOnce()
-		if releaseErr == nil {
-			return
-		}
-		if callbackErr == nil {
-			callbackErr = releaseErr
-			return
-		}
-		callbackErr = errors.Join(callbackErr, releaseErr)
-	}()
-
-	callbackErr = callWithPanicSafetyError(function, deployment)
-
-	return callbackErr
+	return nil
 }
 
-func mapLockAcquireError(ctx context.Context, deployment config.DeploymentDir, err error) error {
+func mapDeploymentLockError(deployment config.DeploymentDir, err error) error {
 	if errors.Is(err, context.Canceled) {
 		return err
 	}
 	if !errors.Is(err, directorymutex.ErrAcquireTimeout) {
 		return err
-	}
-	if ctx == nil {
-		return &deploymentDirectoryLockedError{message: lockUnavailableMessage}
 	}
 
 	if operation := lockedOperationName(deployment); operation != "" {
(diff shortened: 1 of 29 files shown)


internal/runtimeartifacts/manager.go
```diff
@@ -81,149 +153,227 @@ func (m *Manager) Request(ctx context.Context, resourceID string) (string, error
...
-	cache, err := m.readCache()
-	if err != nil {
-		return "", err
-	}
-
-	if entry, ok := cache.Resources[resourceID]; ok {
-		if resolvedPath, err := validateCacheEntry(entry, artifact, target); err == nil {
-			return resolvedPath, nil
+	var resolvedPath string
+	err = m.cache.withExclusiveLock(ctx, func() error {
```
nh-exa, line 157: One concern I have is that we hold the exclusive cache lock while downloading and extracting artifacts ( refresh() is doing all of that, right).
Since directorymutex falls back to a 1s acquire timeout when there’s no deadline on the context, a second deployment running in parallel will often fail while the first one is still warming the cache. That seems especially likely on first-time downloads or slower connections.

I think we should at least give the runtime artifact cache its own longer lock timeout. That still isn't great though. Holding locks for long is never great.
It might be worth avoiding the global lock during the download/extract step: stage the artifact in a unique temp location first, then take the lock only to re-check the cache, move the artifact into place, and update the index.

nh-exa (changes requested):

internal/runtimeartifacts/manager.go
migroble: Increased the timeout to 5 minutes, that should be long enough. If lock contention ever becomes a problem we can implement a smarter locking scheme.

nh-exa (dismissed):
reviewed architecture, looks good.

internal/runtimeartifacts/cache.go
```diff
@@ -0,0 +1,542 @@
...
+			plan, planErr := c.planPartialDownloadCleanup()
+			summary = plan.summary(opts.DryRun)
+			if planErr != nil || opts.DryRun {
+				return planErr
+			}
+
+			return removePartialDownloads(plan)
+		}
+
+		index, _, err := c.readIndex()
```
loaj84, line 270: clean --all bails out if index.json can't be parsed — but that's exactly when you'd want to use it. since it wipes everything anyway, could we just fall back to an empty index instead of erroring on a bad read?

internal/deploy/locking.go
```diff
@@ -65,54 +64,31 @@ func withDeploymentLock(
 		return err
 	}
 
+	wrapped := func(arg any) error {
```
loaj84: deployment's already in scope here, so can't we just close over it instead of passing it through `arg any` and casting it back? also the value, _ := arg.(...) drops the ok, so a wrong type would quietly turn into an empty DeploymentDir instead of failing.

loaj84 (changes requested):

internal/deploy/locking.go
migroble: `arg` is guaranteed to be the right type here because we are passing it a few lines after, but I agree that is probably clearer.

nh-exa (approved):

loaj84 (approved):

PR #144, opened 2026-06-12
Various minor fixes and improvements for Exasol Personal Local

wizenink:
## Description

Follow-up PR for #129 

## Related Issue

Fixes #

## Type of Change

- [x] `feat`: New feature
- [x] `fix`: Bug fix
- [x] `chore`: Maintenance tasks

## Changes Made

Contains the folowing work:

* feat: Use resource manager to download the VM files at build time
* docs: Explicitly document unsafe behaviour on ssh connect
* feat: Use EdDSA keys instead of RSA for local
*  feat: Move multiline strings and scripts to assets

## Testing

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

### DIFF
--- a/assets/infrastructure/local/files/container_shell.sh
+++ b/assets/infrastructure/local/files/container_shell.sh
@@ -1,18 +1,9 @@
-container_name="__LOCAL_DB_CONTAINER_NAME__"
-if ! podman container exists "$container_name"; then
-  container_name="__LOCAL_RUNNER_COMPATIBILITY_DB_CONTAINER_NAME__"
-fi
+container_name="exasol-local-db"
 if ! podman container exists "$container_name"; then
   echo "Exasol Local database container not found" >&2
   podman ps -a >&2
   exit 125
 fi
-
-for shell_path in /bin/bash /usr/bin/bash /bin/sh /usr/bin/sh /bin/ash /usr/bin/ash; do
-  if podman exec "$container_name" "$shell_path" -c 'exit 0' >/dev/null 2>&1; then
-    exec podman exec -it "$container_name" "$shell_path"
-  fi
-done
 
 rootfs=$(podman mount "$container_name") || exit $?
 pid=$(podman inspect "$container_name" --format '{{.State.Pid}}') || exit $?
--- a/assets/local/scripts/container_shell.sh
+++ b/assets/local/scripts/container_shell.sh
@@ -1,0 +1,25 @@
+container_name="__LOCAL_DB_CONTAINER_NAME__"
+if ! podman container exists "$container_name"; then
+  container_name="__LOCAL_RUNNER_COMPATIBILITY_DB_CONTAINER_NAME__"
+fi
+if ! podman container exists "$container_name"; then
+  echo "Exasol Local database container not found" >&2
+  podman ps -a >&2
+  exit 125
+fi
+
+for shell_path in /bin/bash /usr/bin/bash /bin/sh /usr/bin/sh /bin/ash /usr/bin/ash; do
+  if podman exec "$container_name" "$shell_path" -c 'exit 0' >/dev/null 2>&1; then
+    exec podman exec -it "$container_name" "$shell_path"
+  fi
+done
+
+rootfs=$(podman mount "$container_name") || exit $?
+pid=$(podman inspect "$container_name" --format '{{.State.Pid}}') || exit $?
+echo "Exasol Local database container image does not include a shell; using VM shell with container rootfs mounted as working directory."
+echo "Container rootfs: $rootfs"
+cd "$rootfs" || exit $?
+if [ -n "$pid" ] && [ "$pid" != "0" ]; then
+  exec nsenter --target "$pid" --uts --ipc --net /bin/sh
+fi
+exec /bin/sh
--- a/assets/resources/resources.yaml
+++ b/assets/resources/resources.yaml
@@ -25,7 +25,6 @@ exasol-local-runner:
   extract: true
   artifact:
     darwin/arm64:
-      url: https://github.com/exasol/exasol-local-vm/releases/download/v1.0.0-rc1/mac-runner-aarch64.zip
-      sha256: 9f3fdb37a7f219ab66723f2656218817eab71fb13f802e9c5f4cd70992647e2f
-      download_path: mac-runner-aarch64.zip
+      url: https://github.com/exasol/exasol-local-vm/releases/download/v1.0.0-rc2/mac-runner-aarch64.zip
+      sha256: 393b44e41fb98e85a68f74876583c9250d2fdbf9a851e74ab8592713951762f0
       resource_path: launcher
--- a/assets/resources/resources.yaml
+++ b/assets/resources/resources.yaml
@@ -21,3 +21,11 @@ tofu:
       url: https://github.com/opentofu/opentofu/releases/download/v1.11.5/tofu_1.11.5_windows_amd64.tar.gz
       sha256: 69be358054afcc6a84c0ffb917056d21afc3ec5a0e0f66ffd4f498a1763a0fb8
       resource_path: tofu.exe
+exasol-local-runner:
+  extract: true
+  artifact:
+    darwin/arm64:
+      url: https://github.com/exasol/exasol-local-vm/releases/download/v1.0.0-rc1/mac-runner-aarch64.zip
+      sha256: 9f3fdb37a7f219ab66723f2656218817eab71fb13f802e9c5f4cd70992647e2f
+      download_path: mac-runner-aarch64.zip
+      resource_path: launcher
--- a/doc/development.md
+++ b/doc/development.md
@@ -101,7 +101,7 @@ If you prefer to use Go commands directly (or Task is unavailable):
 go generate ./...
 
 # For macOS Apple Silicon builds, stage the Exasol Local runner for embedding.
-# This uses RUNNER_PATH or the runtime artifact configured in resources.yaml.
+# This uses RUNNER_PATH or the runtime artifact configured in assets/resources/resources.yaml.
 go run ./tools/localrunner stage
 
 # Build the binary
--- a/doc/development.md
+++ b/doc/development.md
@@ -86,7 +86,7 @@ GOOS=windows GOARCH=amd64 task build
 GOOS=darwin GOARCH=arm64 task build
 ```
 
-**Note:** The launcher resolves [OpenTofu](https://opentofu.org/) at runtime, so builds no longer need platform-specific OpenTofu downloads. For macOS Apple Silicon builds, `task generate` stages the Exasol Local runner from `RUNNER_PATH`, a cached runner, or the public Exasol Local VM release download.
+**Note:** The launcher resolves [OpenTofu](https://opentofu.org/) at runtime, so builds no longer need platform-specific OpenTofu downloads. For macOS Apple Silicon builds, `task generate` stages the Exasol Local runner from `RUNNER_PATH` or the runtime artifact configured in `assets/resources/resources.yaml`.
 
 ```bash
 GOOS=darwin GOARCH=arm64 task generate RUNNER_PATH=/path/to/launcher
@@ -101,7 +101,7 @@ If you prefer to use Go commands directly (or Task is unavailable):
 go generate ./...
 
 # For macOS Apple Silicon builds, stage the Exasol Local runner for embedding.
-# This uses RUNNER_PATH, a cached runner, or the release download.
+# This uses RUNNER_PATH or the runtime artifact configured in resources.yaml.
 go run ./tools/localrunner stage
 
 # Build the binary
--- a/internal/deploy/local_backend.go
+++ b/internal/deploy/local_backend.go
@@ -21,24 +21,24 @@ import (
 )
 
 const (
-	localSupportedOS           = "darwin"
-	localSupportedArch         = "arm64"
-	localAllowUnsupportedEnv   = "EXASOL_LOCAL_ALLOW_UNSUPPORTED_PLATFORM"
-	localSkipDatabaseWaitEnv   = "EXASOL_LOCAL_SKIP_DB_WAIT"
-	localDefaultCPUCount       = 2
-	localDefaultMemoryMB       = 2048
-	localDefaultDataSizeGB     = 100
-	localDeploymentBackend     = "local"
-	localDeploymentPublicHost  = "127.0.0.1"
-	localSSHUser               = "root"
-	localDBUser                = "sys"
-	localDBPassword            = "exasol"
-	localDBContainerName       = "exasol-local-db"
-	localLegacyDBContainerName = "exasol-nano-db"
-	localManifestFileMode      = 0o600
-	localCPUCountConfigName    = "cpu_count"
-	localMemoryMBConfigName    = "memory_mb"
-	localDataSizeGBConfigName  = "data_size_gb"
+	localSupportedOS                        = "darwin"
+	localSupportedArch                      = "arm64"
+	localAllowUnsupportedEnv                = "EXASOL_LOCAL_ALLOW_UNSUPPORTED_PLATFORM"
+	localSkipDatabaseWaitEnv                = "EXASOL_LOCAL_SKIP_DB_WAIT"
+	localDefaultCPUCount                    = 2
+	localDefaultMemoryMB                    = 2048
+	localDefaultDataSizeGB                  = 100
+	localDeploymentBackend                  = "local"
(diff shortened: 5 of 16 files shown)


assets/local/scripts/container_shell.sh
```diff
@@ -0,0 +1,25 @@
...
+if ! podman container exists "$container_name"; then
+  container_name="__LOCAL_RUNNER_COMPATIBILITY_DB_CONTAINER_NAME__"
+fi
+if ! podman container exists "$container_name"; then
+  echo "Exasol Local database container not found" >&2
+  podman ps -a >&2
+  exit 125
+fi
+
+for shell_path in /bin/bash /usr/bin/bash /bin/sh /usr/bin/sh /bin/ash /usr/bin/ash; do
```
migroble: I'm still not convinced we need to be this paranoid about looking for the shell, I think after seeing that the container exists we can just `podman exec` into it. If the shell in the DB image ever changes, we can update this script to check for the old and the new shells to preserve backwards compatibility.

assets/resources/resources.yaml
```diff
@@ -21,3 +21,11 @@ tofu:
       url: https://github.com/opentofu/opentofu/releases/download/v1.11.5/tofu_1.11.5_windows_amd64.tar.gz
       sha256: 69be358054afcc6a84c0ffb917056d21afc3ec5a0e0f66ffd4f498a1763a0fb8
       resource_path: tofu.exe
+exasol-local-runner:
+  extract: true
+  artifact:
+    darwin/arm64:
+      url: https://github.com/exasol/exasol-local-vm/releases/download/v1.0.0-rc1/mac-runner-aarch64.zip
+      sha256: 9f3fdb37a7f219ab66723f2656218817eab71fb13f802e9c5f4cd70992647e2f
+      download_path: mac-runner-aarch64.zip
```
migroble: You can leave out the `download_path` field, it's only useful in the very weird case where you have a URL for downloading a file which doesn't end in a valid filename.

doc/development.md
```diff
@@ -98,7 +98,7 @@ If you prefer to use Go commands directly (or Task is unavailable):
 go generate ./...
 
 # For macOS Apple Silicon builds, stage the Exasol Local runner for embedding.
-# This uses RUNNER_PATH, a cached runner, or the release download.
+# This uses RUNNER_PATH or the runtime artifact configured in resources.yaml.
```
migroble: I think you should qualify `resources.yaml` as `assets/resources/resources.yaml` since eventually we might want to allow presets to override the `resources.yaml` file.

internal/deploy/local_shell_instructions.tmpl
```diff
@@ -0,0 +1,8 @@
+=== Local Shell Instructions ===
```
migroble: Using templates for big strings is pretty cute, we should do a lot more of that!

internal/deploy/connection_instructions.go
```diff
@@ -127,19 +127,25 @@ To connect using the CLI:
 	instructions += getAdminUIInstructions(connectionDetails)
 
 	if !connectionDetails.ShellSupported {
-		return instructions
+		return instructions, nil
 	}
 
 	if connectionDetails.Backend == localDeploymentBackend {
```
migroble: We might want to consider adding a method to the backends to display their connection info, I don't like switching on backend types, it defeats the purpose of having an interface. That doesn't have to be part of this PR, just worth noting.

internal/deploy/local_backend.go
```diff
@@ -21,24 +21,24 @@ import (
...
+	localDefaultCPUCount                    = 2
+	localDefaultMemoryMB                    = 2048
+	localDefaultDataSizeGB                  = 100
+	localDeploymentBackend                  = "local"
+	localDeploymentPublicHost               = "127.0.0.1"
+	localSSHUser                            = "root"
+	localDBUser                             = "sys"
+	localDBPassword                         = "exasol"
+	localDBContainerName                    = "exasol-local-db"
+	localRunnerCompatibilityDBContainerName = "exasol-nano-db"
```
migroble: I don't think we need to keep backwards compatibility here. This can be removed from here + the script.

nh-exa: do we actually need that?

internal/deploy/local_backend.go
```diff
@@ -21,24 +21,24 @@ import (
...
-	localDBPassword            = "exasol"
-	localDBContainerName       = "exasol-local-db"
-	localLegacyDBContainerName = "exasol-nano-db"
-	localManifestFileMode      = 0o600
-	localCPUCountConfigName    = "cpu_count"
-	localMemoryMBConfigName    = "memory_mb"
-	localDataSizeGBConfigName  = "data_size_gb"
+	localSupportedOS                        = "darwin"
+	localSupportedArch                      = "arm64"
+	localAllowUnsupportedEnv                = "EXASOL_LOCAL_ALLOW_UNSUPPORTED_PLATFORM"
```
nh-exa: is that still needed?

internal/deploy/local_backend.go
```diff
@@ -21,24 +21,24 @@ import (
...
+	localSkipDatabaseWaitEnv                = "EXASOL_LOCAL_SKIP_DB_WAIT"
```
nh-exa: is that still needed?

internal/deploy/connection_instructions.go
```diff
@@ -127,19 +127,25 @@ To connect using the CLI:
...
-		return instructions + `=== Local Shell Instructions ===
-  Local endpoint: ` + displayHostname(connectionDetails) + `
-  Exasol Local database container shell: exasol shell container
-  Host shell (VM OS): exasol shell host
-  Alternative: ` + connectionDetails.SSHCommand + `
-
-Note: exasol destroy deletes the local VM disk/data and launcher-managed share for this deployment.
+		localInstructions, err := renderLocalAssetTemplate(
+			localShellInstructionsTemplateAssetPath,
```
nh-exa: this isn't code just - terminal output. we have lots of that which which isn't in assets. would have been more consistent to leave that in Go code as constant.

e.g. look at "getAdminUIInstructions" that is even worse than this. maybe we should sweep through all code and extract those big text templates as assets?

nh-exa (dismissed):

internal/deploy/local_runtime.go
```diff
@@ -22,33 +22,65 @@ func deployLocalRuntime(
...
+}
+
+func startLocalRuntime(
+	ctx context.Context,
+	deployment config.DeploymentDir,
+	runtimeConfig localRuntimeConfig,
+	out, outErr io.Writer,
+) error {
+	return prepareAndStartLocalRuntime(ctx, deployment, runtimeConfig, out, outErr)
+}
```
migroble: These two functions appear to be exactly the same.

In general, I'm not a big fan of function names with "if" or "and" in the name since that is usually indicative of spaghetti code that doesn't compose well. I think all the functions in this file should be refactored into single-purpose primitives that a higher-level function calls in sequence, instead of that sequence being implicit inside a deep callstack within the primitives.

LLMs really like doing things in super deep callstacks instead of composable primitives, for some reason.

tools/localrunner/main.go
```diff
@@ -57,14 +45,12 @@ func main() {
 
 func run(ctx context.Context, args []string) error {
 	if len(args) == 0 {
-		return errors.New("expected subcommand: placeholder, download, or stage")
+		return errors.New("expected subcommand: placeholder or stage")
```
migroble, line 46: Instead of "stage", perhaps "runner" would be more clear.

tools/localrunner/main.go
```diff
@@ -182,358 +162,87 @@ func resolveRunnerSource(ctx context.Context, config runnerConfig) (string, erro
...
-
-		return false, err
-	}
-	defer func() {
-		_ = source.Close()
-	}()
-
-	dest, err := os.Create(destPath)
-	if err != nil {
+func copyExecutable(sourcePath, targetPath string) (bool, error) {
```
migroble, line 165: Can't this whole function be replaced with an `io.Copy` + the `sameFileContent` check and then inlined?

tools/localrunner/main.go
```diff
@@ -141,29 +122,28 @@ func parseRunnerFlags(name string, args []string) (runnerConfig, error) {
...
+	flags.StringVar(
+		&config.resourceID,
+		"resource-id",
+		defaultResourceID,
+		"Resource ID in resources.yaml",
+	)
 	flags.StringVar(&config.targetPath, "target", defaultGeneratedPath, "Staged runner path")
 	flags.StringVar(&config.runnerPath, "runner-path", "", "Existing runner binary path")
-	flags.StringVar(&config.baseURL, "base-url", "", "Release asset base URL override")
+	flags.StringVar(&config.cacheRoot, "cache-root", "", "Runtime artifact cache root")
```
migroble: Is this needed? If you get rid of it, you can simplify `newResourceManager` a bit.

tools/localrunner/main.go
migroble: I don't think this function provides much value, every environment variable can be passed in as an explicit parameter already.

tools/localrunner/main.go
```diff
@@ -96,17 +82,6 @@ func runPlaceholder(args []string) error {
...
-	if err != nil {
-		return err
-	}
-
-	targetPath := filepath.Join(config.downloadDir, "mac-runner-aarch64")
-
-	return downloadRunner(ctx, config, targetPath)
-}
-
 func runStage(ctx context.Context, args []string) error {
```
migroble: I'm not a huge fan of these `run*` functions, their naming is a bit confusing to me, maybe `preparePlaceholder`/`prepareRunner`?

tools/localrunner/main.go
migroble: The whole `resolveRunnerSource` path could be simplified a lot, take a look at how [`internal/tofu/resource.go`](https://github.com/exasol/exasol-personal/blob/dcfab2ff5910ab4b8d74ed1eed949d45ebb7e50c/internal/tofu/resource.go#L14) does it.

migroble (dismissed):

internal/deploy/connection_instructions.go
migroble: I think templates are the way to go, maybe they should be at `assets/templates` instead of `assets/local/templates`.

assets/local/scripts/container_shell.sh
```diff
@@ -0,0 +1,25 @@
+container_name="__LOCAL_DB_CONTAINER_NAME__"
```
migroble: Should this be in `assets/infrastructure/local/files/`?

internal/deploy/local_backend.go
```diff
@@ -265,40 +265,33 @@ func (b *localBackend) OpenCOSShell(ctx context.Context) error {
...
+	if err != nil {
+		return "", err
+	}
+
+	command = strings.ReplaceAll(command, "__LOCAL_DB_CONTAINER_NAME__", localDBContainerName)
+	command = strings.ReplaceAll(
+		command,
+		"__LOCAL_RUNNER_COMPATIBILITY_DB_CONTAINER_NAME__",
+		localRunnerCompatibilityDBContainerName,
+	)
```
migroble: Instead of string replacements, we could pass the container name as an argument (or even hardcode it).

nh-exa:
You kind of see the impact of using AI here.
All kind of works but if you look at the stuff a bit deeper, you find lots of inefficient architecture and code smells easily. 
This happens with handwritten code a lot too but it happens especially if you dont give your LLM some meta-instructions about architecture goals and code best practices.

So I think it is still worth re-viewing and reconsidering the implementations details as they can have a huge impact on long-term maintainability.

We should do this more often. Refactoring is cheap! And fun!

migroble:
I certainly see the same few patterns appearing in LLM-generated code such as the aforementioned "deep callstack instead of composable primitives". We should consider adding these as coding guidelines for LLMs to avoid having to repeat it to them, I'll try experimenting with that idea in future PRs of mine.

assets/infrastructure/local/files/container_shell.sh
```diff
@@ -0,0 +1,16 @@
...
+
+rootfs=$(podman mount "$container_name") || exit $?
+pid=$(podman inspect "$container_name" --format '{{.State.Pid}}') || exit $?
+echo "Exasol Local database container image does not include a shell; using VM shell with container rootfs mounted as working directory."
+echo "Container rootfs: $rootfs"
+cd "$rootfs" || exit $?
+if [ -n "$pid" ] && [ "$pid" != "0" ]; then
+  exec nsenter --target "$pid" --uts --ipc --net /bin/sh
+fi
+exec /bin/sh
```
migroble, line 16: This could be replaced with a `exec podman exec -it "$container_name" /bin/bash`, the container is guaranteed to have a shell, we don't ever need to fall back.

wizenink: I cannot find any shell on the new container, its a minimal distroless setup:

```
localhost:~# rootfs=$(podman mount exasol-local-db)
localhost:/var/lib/containers/storage/overlay/9da9641b04e9a7cb2c4445b59ac9d9b0f52e1dfac3693e9aee8b24415bd3d198/merged# find "$rootfs" -type f -name bash
localhost:/var/lib/containers/storage/overlay/9da9641b04e9a7cb2c4445b59ac9d9b0f52e1dfac3693e9aee8b24415bd3d198/merged# find "$rootfs" -type f -name sh
localhost:/var/lib/containers/storage/overlay/9da9641b04e9a7cb2c4445b59ac9d9b0f52e1dfac3693e9aee8b24415bd3d198/merged# find "$rootfs" -type f -name ash
localhost:/var/lib/containers/storage/overlay/9da9641b04e9a7cb2c4445b59ac9d9b0f52e1dfac3693e9aee8b24415bd3d198/merged# find "$rootfs" -type f -name zsh
localhost:/var/lib/containers/storage/overlay/9da9641b04e9a7cb2c4445b59ac9d9b0f52e1dfac3693e9aee8b24415bd3d198/merged# find "$rootfs" -type f \( -name sh -o -n
ame bash -o -name ash -o -name zsh \) 2>/dev/null
localhost:/var/lib/containers/storage/overlay/9da9641b04e9a7cb2c4445b59ac9d9b0f52e1dfac3693e9aee8b24415bd3d198/merged#
```

internal/deploy/local_backend.go
wizenink: supportedOS and supportedArch, yes. The other ones were test-used stuff that leakeds into runtime, removed

internal/deploy/connection_instructions.go
wizenink: IMO templates are absolutely the way to go, makes the go code cleaner. Noting moving extra things to templates for a future PR

assets/resources/resources.yaml
wizenink: resolved

internal/deploy/connection_instructions.go
wizenink: noted for a future follow-up PR

internal/deploy/local_backend.go
wizenink: changed the container name on the VM side of this PR, backward compatibility dropped now

assets/infrastructure/local/files/container_shell.sh
migroble: Does it even make sense to provide a shell for the container then? What would the user be able to do with it?

internal/deploy/local_shell_instructions.tmpl
migroble: It seems we have already done more of that since this comment. This file should be next to the source that uses it and have the `.tmpl` extension to match #148.

internal/runtimeartifacts/download.go
```diff
@@ -5,6 +5,7 @@ package runtimeartifacts
 
 import (
```
migroble, line 6: You can get rid of adding `.zip` support here, it's being added as part of #152, which should be merged by Monday (hopefully).

assets/infrastructure/local/files/container_shell.sh
wizenink: well, its a minor QOL command that basically mounts the podman NS and gets you into the directory. Not super useful, but the alternative was to error out and tell the user that on local deployments, `shell container` is not supported, and this way you skip a `podman mount`

internal/runtimeartifacts/download.go
wizenink: Let's merge this and get the deduplication in a follow-up PR after #152 is merged

nh-exa (dismissed):

nh-exa (approved):

PR #145, opened 2026-06-12
chore: optimize launcher binary size

nh-exa:
## Description

Reduce raw Exasol launcher binary size by applying low-effort Go build optimizations to local Task builds and GoReleaser release builds. The change also documents the accepted and rejected techniques and records the decision in OpenSpec.

Observed raw binary reductions:

| Platform | Before | After | Reduction |
|---|---:|---:|---:|
| Linux amd64 | 21,709,121 B | 13,709,496 B | 7,999,625 B, 36.9% |
| Linux arm64 | 20,260,292 B | 12,583,096 B | 7,677,196 B, 37.9% |
| macOS amd64 | 22,157,824 B | 14,019,904 B | 8,137,920 B, 36.7% |
| macOS arm64 | 20,865,762 B | 12,948,706 B | 7,917,056 B, 37.9% |
| Windows amd64 | 22,243,328 B | 14,204,928 B | 8,038,400 B, 36.1% |

## Related Issue

No related issue ID.

## Type of Change

- [ ] `feat`: New feature
- [ ] `fix`: Bug fix
- [x] `chore`: Maintenance tasks

## Changes Made

- Applied `CGO_ENABLED=0`, `-trimpath`, `-gcflags=all=-l`, and `-ldflags="-s -w ..."` to Task launcher builds.
- Applied matching binary size flags to GoReleaser builds.
- Added concise binary size optimization documentation with effort/risk tradeoffs and references.
- Added OpenSpec artifacts for `optimize-go-binary-size`.

## Testing

### Test Details

- `env GOCACHE=/tmp/go-build-cache task build`
- `env GOCACHE=/tmp/go-build-cache task build-windows`
- `task tests-unit`

`goreleaser` is not installed in the local environment, so native GoReleaser validation was not run.

## Checklist

- [ ] Added/updated tests
- [x] All tests pass locally

## Additional Notes

Higher-risk binary size techniques such as executable packing, alternative compiler toolchains, and dependency replacement for size alone were intentionally rejected due to signing, debugging, release, and maintenance tradeoffs.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -93,19 +93,22 @@ tasks:
     desc: Build the binary
     deps:
       - generate
+    vars:
+      DEBUG_BUILD: '{{default "false" .DEBUG_BUILD}}'
     env:
       CGO_ENABLED: "0"
     cmds:
       - |
         version=$(git describe --tags --abbrev=0) # First character is "v".
         go build -o {{.BUILD_DIR_NAME}}/{{.BINARY_NAME}} \
-          -trimpath \
-          -gcflags=all=-l \
-          -ldflags "-s -w -X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
+          {{if eq .DEBUG_BUILD "true"}}-gcflags="all=-N -l"{{else}}-trimpath \
+          -gcflags=all=-l{{end}} \
+          -ldflags "{{if ne .DEBUG_BUILD "true"}}-s -w {{end}}-X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
 
   build-windows:
     desc: Cross-compile the Windows binary (defaults to GOARCH=amd64)
     vars:
+      DEBUG_BUILD: '{{default "false" .DEBUG_BUILD}}'
       TARGET_GOARCH: '{{default "amd64" .GOARCH}}'
       OUTPUT_NAME: '{{.COMMAND_NAME}}.exe'
     env:
@@ -116,9 +119,9 @@ tasks:
       - |
         version=$(git describe --tags --abbrev=0) # First character is "v".
         go build -o {{.BUILD_DIR_NAME}}/{{.OUTPUT_NAME}} \
-          -trimpath \
-          -gcflags=all=-l \
-          -ldflags "-s -w -X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
+          {{if eq .DEBUG_BUILD "true"}}-gcflags="all=-N -l"{{else}}-trimpath \
+          -gcflags=all=-l{{end}} \
+          -ldflags "{{if ne .DEBUG_BUILD "true"}}-s -w {{end}}-X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
 
 
   deps-update-golang:
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -93,11 +93,15 @@ tasks:
     desc: Build the binary
     deps:
       - generate
+    env:
+      CGO_ENABLED: "0"
     cmds:
       - |
         version=$(git describe --tags --abbrev=0) # First character is "v".
         go build -o {{.BUILD_DIR_NAME}}/{{.BINARY_NAME}} \
-          -ldflags "-X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
+          -trimpath \
+          -gcflags=all=-l \
+          -ldflags "-s -w -X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
 
   build-windows:
     desc: Cross-compile the Windows binary (defaults to GOARCH=amd64)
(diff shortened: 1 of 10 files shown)


openspec/changes/optimize-go-binary-size/specs/binary-size-optimization/spec.md
```diff
@@ -0,0 +1,16 @@
+## ADDED Requirements
+
+### Requirement: Optimized launcher binary builds
```
migroble: I'm not sure if these two items can be considered requirements, they feel more like "implementation guidelines", they make sense as a mechanism to communicate the task to an LLM, but I don't know if they should be persisted in the repo.

Taskfile.yml
```diff
@@ -93,11 +93,15 @@ tasks:
...
+    env:
+      CGO_ENABLED: "0"
     cmds:
       - |
         version=$(git describe --tags --abbrev=0) # First character is "v".
         go build -o {{.BUILD_DIR_NAME}}/{{.BINARY_NAME}} \
-          -ldflags "-X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
+          -trimpath \
+          -gcflags=all=-l \
+          -ldflags "-s -w -X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
```
migroble: It might be nice to have a variable to keep the debugging info in case we need to use a debugger without having to modify the Taskfile.

migroble:
Those are some pretty substantial gains, we should probably apply them to the VM runner as well.

Do you have a breakdown of how much each optimization reduces the binary size? I doubt any of them will increase the size, but that's not totally unheard of in C/C++, so it doesn't hurt to check.

Taskfile.yml
nh-exa: i'll add that

openspec/changes/optimize-go-binary-size/specs/binary-size-optimization/spec.md
nh-exa: the fact that we do some optimization is a requirement worth keeping. I'll adjust things a bit so this keep it high-level

Taskfile.yml
```diff
@@ -111,8 +124,16 @@ tasks:
...
+        if [ "{{.DEBUG_BUILD}}" = "true" ]; then
+          go build -o {{.BUILD_DIR_NAME}}/{{.OUTPUT_NAME}} \
+            -gcflags="all=-N -l" \
+            -ldflags "-X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
+        else
+          go build -o {{.BUILD_DIR_NAME}}/{{.OUTPUT_NAME}} \
+            -trimpath \
+            -gcflags=all=-l \
+            -ldflags "-s -w -X main.CurrentLauncherVersion=${version:1}" {{.MAIN_PATH_REL}}
+        fi
```
migroble: This should look cleaner using Taskfile's conditionals like we do for the `tests-unit` task, for example.

openspec/changes/optimize-go-binary-size/specs/binary-size-optimization/spec.md
migroble: As requirements, they can't be considered requirements of the product, but rather of the build system. And I don't think mixing the two is a good idea, it only adds noise.

As goals of the product (i.e. "the launcher should be as small as reasonably possible") they would make sense, but they would have to have such weak language as to be nearly useless because in order to truly minimize the size of the executable, we simply would not be using Go or, worse yet, we would have to define "reasonably".

As a prompt for LLMs that just mirrors the aesthetic of requirements for the purposes of "semantic anchoring", I see little value in adding them since the only part this affects is the compilation process and, with this PR, that will already be implemented. If, in the future, we ask an LLM to try to further optimize the size of the executable, these requirements won't really help it.

I'm not really sure these requirements are providing anything useful in the long term -- I see them as a transient artifact of using OpenSpec which we can safely discard.

nh-exa: Which makes it actually important to keep. If i refactor this aspect (using openspec or not) in the future, this might easily get lost otherwise. Requirements are not only functional. There are also non-functional requirements. And this is one that is important. Our CPO asked for it explicitly even.

Taskfile.yml
nh-exa: done

openspec/changes/optimize-go-binary-size/specs/binary-size-optimization/spec.md
migroble: LLMs tend to be extremely conservative about preserving backwards-compatibility when changing behavior, often far too much, so I'm not too concerned about that. Additionally, issues like that would get caught in code reviews as well, just like if a developer had changed it.

I think we should only keep product requirements, the build system is trivial enough in Go that we can avoid that noise.

migroble: Our CPO also explicitly asked for updates to the documentation, but it wouldn't make sense to add requirements that the documentation shall include certain information.

I do agree it is important enough to warrant some documentation, but I think `doc/binary_size.md` should be plenty. That document already provides a pretty good breakdown of what we are doing, why, and alternatives considered.

migroble (approved):

PR #153, opened 2026-06-18
feat: Adjust local infra memory defaults

loaj84:
## Description

This PR makes local VM memory default to about 50% of detected host memory on macOS, enforces minimum memory requirements, and updates the related tests and docs.

## Related Issue

[SPOT-31128](https://exasol.atlassian.net/browse/SPOT-31128)

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- change macOS local VM memory default to about 50% of detected host memory
- reject user-configured local VM memory below 4096 MB
- reject macOS local deployments when detected host memory is below 8192 MB
- simplify local host-memory detection to a darwin-only path
- update local backend unit tests and local integration coverage for the new memory behavior
- update OpenSpec docs and README.md to reflect the new local memory behavior

## Testing

### Test Details

Manually tested the `exasol` launcher binary after integrating the runner artifact built by CI from the `exasol-local-vm` repository.

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

## Additional Notes

exasol-local-vm PR - https://github.com/exasol/exasol-local-vm/pull/22

Dev Testing Logs:  [SPOT-31128_dev_testing_logs_1.txt](https://github.com/user-attachments/files/29161276/SPOT-31128_dev_testing_logs_1.txt)

[SPOT-31128]: https://exasol.atlassian.net/browse/SPOT-31128?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


Gregory-Bull-Exasol (dismissed):
According to Ruslan we only need 512 mb for shared memory now (he fixed something in the container)

His latest recommendations for the container are:

`--security-opt unmask=ALL --shm-size=512mb --pids-limit=-1`

with restart policy set to "always". I was planing to add this but here is as good a place as any.

cmd/exasol/init.go
```diff
@@ -179,6 +179,7 @@ func runInitForFreshDeployment(
 		return err
 	}
 	addTerminalNotice(deploy.EulaNoticeText)
+	addTerminalNotice(deploy.LocalInitMemoryNotice(deployment))
```
migroble: I think this might be more readable if the memory check is performed here or if `deploy.LocalInitMemoryNotice` returned `(message string, hasLowMemory bool)`, otherwise it reads as if we are always adding the memory notice.

internal/deploy/local_backend.go
```diff
@@ -351,42 +358,93 @@ type localRuntimeConfig struct {
...
-		memoryMB:   localDefaultMemoryMB,
+		memoryMB:   memoryMB,
 		dataSizeGB: localDefaultDataSizeGB,
 	}
 }
 
+func detectLocalHostMemoryMB(ctx context.Context) int {
+	if runtime.GOOS != localSupportedOS {
+		return 0
+	}
```
migroble: I think the platform-specific parts should be moved into a separate helper function and this agnostic function should just switch on `runtime.GOOS` to call the appropriate function and panic on the unimplemented variants.

A generic `GetTotalMemoryMB` function would probably fit better into `utils`.

cmd/exasol/init.go
nh-exa: can't say I like this too much for architecture.
Backends have a `ValidateEnvironment()" method that you could use for this check instead of reaching into the internals from the cobra command layer. The main problem here is that wee now try to print text warning based on some internal state. We could just used the logging system for that. it can print warnings from everywhere. otherwise, we somehow need to transport a "warning" out of the internals to the cmd layer via some return value.

nh-exa (changes requested):

internal/deploy/local_backend.go
```diff
@@ -233,10 +239,11 @@ func localConfigVariableDefinitions(
...
 func ensureLocalManifestConfig(
+	ctx context.Context,
 	manifest *presets.InfrastructureManifest,
 ) *presets.InfrastructureLocal {
 	if manifest.Local == nil {
-		defaults := defaultLocalRuntimeConfig()
+		defaults := defaultLocalRuntimeConfig(detectLocalHostMemoryMB(ctx))
 		manifest.Local = &presets.InfrastructureLocal{
 			CPUCount:   defaults.cpuCount,
 			MemoryMB:   defaults.memoryMB,
```
migroble, line 294: I think, after this PR, the `MemoryMB` field is only used in tests and can be removed (unless you change it to a percentage as suggested).

internal/deploy/local_backend.go
nh-exa: agreed

internal/deploy/local_backend.go
```diff
@@ -351,42 +358,93 @@ type localRuntimeConfig struct {
...
 		dataSizeGB: localDefaultDataSizeGB,
 	}
 }
 
+func detectLocalHostMemoryMB(ctx context.Context) int {
+	if runtime.GOOS != localSupportedOS {
+		return 0
+	}
+
+	output, err := exec.CommandContext(
```
nh-exa: is chatgpt lying to me here or doesn't this look better?

```
package main

import (
	"fmt"
	"golang.org/x/sys/unix"
)

func main() {
	// Query the macOS kernel for the total hardware memory size
	totalBytes, err := unix.SysctlUint64("hw.memsize")
	if err != nil {
		fmt.Println("Error retrieving memory size:", err)
		return
	}

	// Convert bytes to Gigabytes (GB)
	totalGB := float64(totalBytes) / (1024 * 1024 * 1024)

	fmt.Printf("Total Memory: %d bytes\n", totalBytes)
	fmt.Printf("Total Memory: %.2f GB\n", totalGB)
}
```

cmd/exasol/init.go
loaj84: I considered a few alternatives, but none felt clearly better here.

ValidateEnvironment() is intended for hard failures and returns an error, whereas this is just a non-fatal recommendation. Logging would make it log-level dependent and show it as a diagnostic rather than a user-facing notice, and it could be missed entirely if the user runs with a higher log level like --log-level error.

Returning a value from InitDeployment() is cleaner architecturally, but it would require changing the function signature and ripple through install.go and several tests without changing the actual behavior.

Since the manifest read and threshold check are already encapsulated in deploy, I'd lean toward keeping the current approach.

loaj84: Addressed

internal/deploy/local_backend.go
loaj84: This is still used in production. --memory-mb / config set memory_mb persists the value via applyLocalConfigOverrides, and resolveLocalRuntimeConfig reads it back for sizing as well as config get.

I think the test-only reference you're seeing is localruntime.Config.MemoryMB, which belongs to a different struct than the one being discussed here.

internal/deploy/local_backend.go
loaj84: The ChatGPT suggestion is solid and it’s actually what I had originally. The problem is that `x/sys/unix` doesn't compile on Windows, so we'd also need to use `x/sys/windows`. Since Go doesn't support OS-specific imports within the same file, that would mean splitting the implementation into multiple build-tagged files just for reading total system memory, which I didn't really like.

To avoid that extra complexity, I went with the `exec` approach instead. It shells out to the native OS command and lets us keep everything in a single cross-platform file.

loaj84:
> According to Ruslan we only need 512 mb for shared memory now (he fixed something in the container)
> 
> His latest recommendations for the container are:
> 
> `--security-opt unmask=ALL --shm-size=512mb --pids-limit=-1`
> 
> with restart policy set to "always". I was planing to add this but here is as good a place as any.

Addressed in exasol-local-vm PR -- https://github.com/exasol/exasol-local-vm/pull/22

assets/infrastructure/local/infrastructure.yaml
```diff
@@ -4,7 +4,6 @@ backend: local
 
 local:
   cpuCount: 2
-  memoryMB: 2048
```
migroble, line 7: I would like to keep this customizable via the presets. Would it be hard to change this from MB to a percentage value and get rid of `hostMemoryDefaultDivisor`?

migroble: (Forgot to post this comment as part of my review)

internal/deploy/local_backend.go
loaj84: addressed

assets/infrastructure/local/infrastructure.yaml
loaj84: Modifying MB to percentage has trade-offs on both sides. A floor (≥50%) helps prevent under-allocation, but on smaller hosts a higher percentage can leave too little for the OS (e.g., 75% of an 8 GB machine leaves only ~2 GB).

To make this safe, we’d need both a floor and a ceiling, and likely a host headroom guard as well which adds quite a few constraints for a knob that mainly helps large hosts.

The fixed 50% split is simple and works reasonably well across host sizes. I’d lean towards keeping it as-is for now, and we can revisit a bounded percentage later if there’s a real need.

internal/deploy/local_backend.go
nh-exa: i still like the multiple files option better. Anything that can be solved at build-time easily should be solved at build-time. It's the same argument as for static types or borrow-checkers.

cmd/exasol/init.go
nh-exa: You don't need to return a value as I explained. I don't think a warning is going  to be unnoticed. We can even make in an error in the log.
A solution which violates architectural boundaries/layers should be avoided. It's much worse than some unclean code and bound to create problems later.

loaj84: addressed.

internal/deploy/local_backend.go
loaj84: addressed.

assets/infrastructure/local/infrastructure.yaml
migroble: Seems reasonable. What do you think about changing `hostMemoryDefaultDivisor` to a percentage value so it's a bit more readable?

internal/deploy/local_backend.go
migroble: Ah, you're right, you can ignore this comment then.

assets/infrastructure/local/infrastructure.yaml
loaj84: makes sense. Addressed.

migroble (dismissed):
Looks good to me, I think all the comments have been addressed.

nh-exa (dismissed):

nh-exa (approved):

PR #157, opened 2026-06-23
fix: exasol help now shows correct cli args for presets

Gregory-Bull-Exasol:
Previously, `exasol help <preset>` would not show the preset's arguments from `infrastructure.yaml` (`--help` did work correctly)

Now, the output includes the preset's args.


nh-exa (dismissed):
need to make the linter happy

nh-exa (approved):

migroble (approved):

PR #161, opened 2026-06-23
feat: CI testing for exasol local

Gregory-Bull-Exasol:
This PR contains:
- enables some of the pre-existing e2e tests to run for exasol local.
- increases timeout and catches errors for creating reusable deployments in the deployment tests. This increases reliability and prevents some orphaned resources.

These tests are run on the self-host mac instance, which requires some additional work to install dependencies due to reduced permissions compared to github's own containers.

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -202,6 +202,21 @@ tasks:
       - |
         poetry run pytest -vv --exasol-path={{.BUILD_DIR}}{{.SEP}}{{.BINARY_NAME}} --infra={{.INFRA}} {{if .STACKIT_PROJECT_ID}}--stackit-project-id={{.STACKIT_PROJECT_ID}}{{end}} -m "installation_e2e" {{.TESTS_DIR}}{{.SEP}}tests{{.SEP}}deployment
 
+  tests-deployment-local:
+    desc: Run local VM deployment tests (macOS Apple Silicon only)
+    deps: [tests-setup]
+    dir: '{{.TESTS_DIR}}'
+    cmds:
+      - poetry run pytest -vv --exasol-path={{.BUILD_DIR}}{{.SEP}}{{.BINARY_NAME}} --infra=local -m "local_e2e" {{.TESTS_DIR}}{{.SEP}}tests{{.SEP}}deployment
+
+  run-deployment-tests:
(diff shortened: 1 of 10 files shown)


nh-exa:
I'd like to see some context for this PR in the description so I can get an idea of what you trying to do before I look into the code

Taskfile.yml
```diff
@@ -202,6 +202,21 @@ tasks:
       - |
         poetry run pytest -vv --exasol-path={{.BUILD_DIR}}{{.SEP}}{{.BINARY_NAME}} --infra={{.INFRA}} {{if .STACKIT_PROJECT_ID}}--stackit-project-id={{.STACKIT_PROJECT_ID}}{{end}} -m "installation_e2e" {{.TESTS_DIR}}{{.SEP}}tests{{.SEP}}deployment
 
+  tests-deployment-local:
+    desc: Run local VM deployment tests (macOS Apple Silicon only)
+    deps: [tests-setup]
+    dir: '{{.TESTS_DIR}}'
+    cmds:
+      - poetry run pytest -vv --exasol-path={{.BUILD_DIR}}{{.SEP}}{{.BINARY_NAME}} --infra=local -m "local_e2e" {{.TESTS_DIR}}{{.SEP}}tests{{.SEP}}deployment
```
nh-exa, line 213: is "tests-deployment" still running everything - including local?

nh-exa: should we have a "tests-deployment-cloud"?

Taskfile.yml
```diff
@@ -202,6 +202,21 @@ tasks:
...
+
+  run-deployment-tests:
+    desc: Trigger the deployment tests workflow on the current pushed commit (`task run-deployment-tests SUITE=local`, `task run-deployment-tests OS=macos-latest`)
```
nh-exa: I think "github:trigger-deployment-tests" would be a better name? maybe to verbose for a name.
But I would like make clear where these tests are running (and that it isn't the local system).

nh-exa (changes requested):

nh-exa (dismissed):

Taskfile.yml
Gregory-Bull-Exasol: https://github.com/exasol/exasol-personal/actions/runs/28037006030 this might make the structure more clear

nh-exa (approved):

PR #172, opened 2026-07-01
Gb.reconcile statefile with vm runner state

Gregory-Bull-Exasol:
Related `exasol-local-vm` side PR: https://github.com/exasol/exasol-local-vm/pull/30

Uses the new mac-runner status socket command to give accurate status and reliable recovery after an unclean local VM shutdown (e.g. SIGKILL).

**Problem**
If the local VM daemon was killed without going through the launcher's stop command, the workflow state file was left as WorkflowStateRunning. This caused two bad outcomes:

`status` reported database_connection_failed (misleading — the VM is simply not running, not failing)
`start` was blocked by the permit check, since WorkflowStateRunning is not a permitted start state

**Changes**
`internal/localruntime/runtime.go`

Adds VMStatus type, a public Status() function that runs mac-runner status and parses its JSON output, and a private runnerCommandWithOutput() helper for commands whose stdout must be captured rather than streamed.

`internal/deploy/local_runtime.go`

`getLocalVMStatus` — thin call-through to localruntime.Status, following the same pattern as stopLocalRuntime → localruntime.Stop.
`reconcileLocalVMState` — called by `start` and `stop` under the exclusive lock before their permit checks. If the workflow state is WorkflowStateRunning but the VM socket reports the daemon is not running, corrects the state to WorkflowStateStopped. Only reconciles in this one direction; Stopped → Running is left as an explicit error.
`isLocalDeployment` — reads the infrastructure manifest to determine backend type; used to no-op both of the above on non-local deployments.

`internal/deploy/deploymentControl.go`

Calls reconcileLocalVMState at the top of Start and Stop, before WorkflowStatePermitsStart/WorkflowStatePermitsStop. This allows start to succeed after an unclean shutdown without any changes to the permit functions.

`internal/deploy/status.go`

In the WorkflowStateRunning + checkConnection=true branch, checks VM liveness via the socket before attempting the DB probe. If the daemon is not running, short-circuits with StatusStopped and a "Run start..." message, avoiding a slow connection timeout against a dead VM. GetStatus remains a pure read.

`tests/tests/deployment/test_local_deployment.py`

Adds `test_status_after_forceful_vm_kill`: deploys a local instance, kills the VM daemon with SIGKILL, and asserts status reports stopped; then runs start and asserts status reports `database_ready`.

### DIFF
--- a/.github/scripts/build-binaries.sh
+++ b/.github/scripts/build-binaries.sh
@@ -1,0 +1,55 @@
+#!/bin/bash
+set -euo pipefail
+
+BRANCH=$(git rev-parse --abbrev-ref HEAD)
+if [[ "$BRANCH" == "HEAD" ]]; then
+    echo "ERROR: Cannot trigger a CI run from a detached HEAD — check out a branch first"
+    exit 1
+fi
+
+LOCAL_SHA=$(git rev-parse HEAD)
+REMOTE_SHA=$(git rev-parse "origin/${BRANCH}" 2>/dev/null) || {
+    echo "ERROR: Branch '${BRANCH}' has not been pushed to origin"
+    exit 1
+}
+
+if [[ "$LOCAL_SHA" != "$REMOTE_SHA" ]]; then
+    echo "ERROR: Local commit ${LOCAL_SHA:0:7} has not been pushed (origin/${BRANCH} is at ${REMOTE_SHA:0:7})"
+    exit 1
+fi
+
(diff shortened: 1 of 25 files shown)


wizenink (approved):

Gregory-Bull-Exasol:
superseded by https://github.com/exasol/exasol-personal/commit/45934ed9c8f5f1d7cf4b70a97180bc3f76013aef

PR #198, opened 2026-07-10
fix: make personal installer POSIX shell compatible

nh-exa:
## Description

Make the Linux/macOS personal installer run with a POSIX-compatible shell instead of requiring Bash. This helps users whose environment invokes the installer via `sh` or another non-Bash shell implementation, including stricter POSIX shells such as `dash`.

## Related Issue

N/A

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- Changed `scripts/personal-installer.sh` shebang from Bash to POSIX `sh`.
- Replaced Bash-only `pipefail` strict mode with portable `set -eu`.
- Kept installer behavior unchanged otherwise.

## Testing

### Test Details

- Ran POSIX shell syntax check with `sh -n scripts/personal-installer.sh`.
- Ran Bash syntax check with `bash -n scripts/personal-installer.sh`.
- Ran ShellCheck in POSIX shell mode with `shellcheck -s sh scripts/personal-installer.sh`.
- Ran an offline mocked installer execution via `sh scripts/personal-installer.sh` with a fake `curl`, verifying the target `exasol` binary was created and marked executable.

## Checklist

- [ ] Added/updated tests
- [ ] All tests pass locally

## Additional Notes

-

### DIFF
--- a/scripts/personal-installer.sh
+++ b/scripts/personal-installer.sh
@@ -1,5 +1,5 @@
-#!/usr/bin/env bash
-set -euo pipefail
+#!/bin/sh
+set -eu
 
 S3_URL="x-up.s3.eu-west-1.amazonaws.com"
 BASE_URL="https://$S3_URL/releases/exasol-personal"


moben (approved):
I _think_ there's nothing bash specific here but I basically never target POSIX sh myself, only bash...

Alternative would be to have the script execute bash itself, probably needs some hacks to support both stdin and files. Or detect this and exit.

If we want to support POSIX sh I'd suggest running `shellcheck --shell=sh` on it, there's probably an action we can just use.

Otherwise this is the sort of thing that's liable to accidentally break again whenever someone touches this

migroble (approved):

migroble:
I agree adding a task to lint shell scripts with shellcheck is a good idea in general.

> If we want to support POSIX sh I'd suggest running shellcheck --shell=sh

shellcheck automatically checks for bashisms when the shebang is `#!/bin/sh`

PR #226, opened 2026-07-21
feat: Upgrade local VM runner on start for minor and patch versions

wizenink:
## Description
Makes exasol personal auto-upgrade the runner on patch/minor version bumps, while keeping major version bumps untouched and gated behind a warning message.

Requires:
- https://github.com/exasol/exasol-local-vm/pull/46
- https://github.com/exasol/exasol-local-vm/pull/47
## Related Issue

Fixes #

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Checks runner version for patch/minor bumps and upgrades the extracted binary only on `start`
- exposes `EXASOL_LOCAL_FORCE_RUNNER_RECONCILIATION` to allow updating a deployment's runner with an unversioned asset.

## Testing

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

### DIFF
--- a/internal/localruntime/runtime.go
+++ b/internal/localruntime/runtime.go
@@ -13,6 +13,7 @@ import (
 	"errors"
 	"fmt"
 	"io"
+	"log/slog"
 	"os"
 	"os/exec"
 	"path/filepath"
@@ -21,6 +22,7 @@ import (
 	"syscall"
 	"time"
 
+	"github.com/blang/semver/v4"
 	"github.com/exasol/exasol-personal/assets/localruntimebin"
 	"github.com/exasol/exasol-personal/internal/config"
 	"github.com/exasol/exasol-personal/internal/util"
@@ -42,6 +44,8 @@ const (
 	privateFileMode    = 0o600
 	executableFileMode = 0o700
 	maxTCPPort         = 65535
+	// Internal escape hatch for development with embedded runners that predate version reporting.
+	forceRunnerReconciliationEnv = "EXASOL_LOCAL_FORCE_RUNNER_RECONCILIATION"
 )
 
 type Config struct {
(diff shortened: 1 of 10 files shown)


internal/localruntime/runtime.go
```diff
@@ -301,6 +309,204 @@ func (runtime *localRuntime) ensureRunnerExecutable() error {
...
+
+	return reconcileRunner(ctx, runtime.paths.RunnerPath, runtime.embeddedRunner)
+}
+
+func reconcileRunner(ctx context.Context, targetPath string, embeddedRunner []byte) error {
+	if err := os.MkdirAll(filepath.Dir(targetPath), dirMode); err != nil {
+		return fmt.Errorf("failed to create local runner target directory: %w", err)
+	}
+
+	candidatePath, cleanup, err := writeRunnerCandidate(targetPath, embeddedRunner)
```
nh-exa, line 336: I dont think I like that too much.
Why isn't this just handled by the cache as normal? Is it because our cache can't hold different version of the same resource? in that case, we need to conceptually rethink the "resource manager" a bit.

I could have two different version of Exasol Personal on my system very easily. They should not fight over a single cache entry for this.

also, the "exasol-local-runner" cache entry has `resource_path: launcher` - which is all kind of confusing. Lets choose a better one. @migroble

nh-exa (changes requested):

internal/localruntime/runtime.go
migroble: The resource cache can have multiple versions of the same resource. We identify each entry by the hash of a bunch of fields including the URL and SHA256, so if those change, it's considered a new entry.

migroble: Keep in mind that the VM runner doesn't use the resource management machinery at runtime, only at build-time. The VM runner is embedded in the binary and extracted on demand.

At first glance, I'm guessing this code is determining whether the runner in the deployment directory is stale and overriding it if so.

If we want to put the VM runner in the cache (for the same reasons we did that with tofu), we could implement an "embedded" resource source or something similar, but that might be a bit convoluted.

nh-exa: Looks like I shouldn't do code reviews when i have a headache.
I did assume the resource manager and cache would be involved even at runtime.
But of course, if it's embedded the resource manager likely isn't involved. 

---
Still, extracting the embedded launcher it into some special different path and performing checks whether it could and should update, every time we run `exasol start`, seems like a job a the resource manager could do much better and more efficiently.

Like intuitively, I think I'd extract it into the cache once and not every time. And we'd use it from there afterwards instead of writing it into each deployment too directory. Because it's not really specific for a deployment, is it?
And then this would also spare us from write a "candidate" for upgrade into some temp dir every single time that `start` command is used. 
But ok, that's not how it is at the moment. 

And as is, `ensureRunnerExecutable` is just forwarding to `writeEmbeddedRunner` and i guess the `os.Stat(targetPath)` call should be in `ensureRunnerExecutable` and not in the `writeEmbeddedRunner`.

wizenink: Well, looks like github decided to not forward this to me, sorry for the delay.
I agree that we could and should use the resource manager to cache runtime assets aswell, but maybe that can come in a follow-up PR expanding the resource manager to runtime usecases. 
That way, we could f.ex not have vm runners taking space in each deployment dir, and just use the cached, corresponding runner.

nh-exa (approved):
Approved.

But @migroble Is going to work on a followup PR to improve here. I don't think it should stay like this,

PR #242, opened 2026-07-27
refactor: Move ownership of podman to exasol-personal SPOT-31797

moben:
## Description

**Status:**
- very draft, I haven't read all of it yet which I definitely need to do, but the general design is decent I think.
- Mixes a few different changes that we might want to split up.
- Ran out of credits during the last commit, so might not even build right now until I do some cleanup.
- Need to rebase.

Completely written by codex.

This would move ownership of anything database related and podman images / containers to `exasol-personal`, so `exasol-local-vm` is just a small vm provider for MacOS.

My main motivations are:
- This way the main functionality isn't split across repositories and features like e.g. SLC support would only need changes in exasol-personal.
- With the Windows support and future Linux support, the exasol-local-vm doesn't actually run a vm and basically becomes a wrapper around podman, but we can't even share a lot of code between the podman setups because they run in different environments. Moving the boundary to be "exasol-personal runs podman on infrastructure" makes code sharing easier.
- Looking ahead to ideas like adding additional sidecar containers we'd either split more functionality between exasol-personal and exasol-local-vm or bundle everything in the vm.
    - Either option would then prevent us making the container setup user-serviceable like in cloud presets. (to be clear, this PR also isn't user-servicable but I argue it would be easier)
- While exasol-local-vm can currently in theory be used standalone to run Exasol in a vm, I don't think anyone would ever use it standalone. If you wanted standalone without exasol-personal you'd just use a container manager and run nano directly.

Currently this is a net reduction of ~5000 lines across both repositories but as noted above there's possibly some stuff missing and some of the changes here could also be implemented in exasol-local-vm and would reduce code there. So this should mainly be judged on if we think going in this direction is helpful. I also know I need to review a bit more what got dropped in the move, there's definitely some helpers in exasol-local-vm for Windows podman / winget that I want to move over and preserve comments.

One argument against this is of course that it adds a bunch of platform-specific code to exasol-personal which currently doesn't have much platform specific code. If exasol-personal should own the container setup (and I'm arguing that it should), then I don't see a good way around that. We could introduce a thin podman wrapper that's more aligned with the actual podman cli but does exasol-local-vm / podman-wsl2 machine / direct-linux podman? Might be helpful if we want to allow user customization of local deployments.

### Details

- There should be no new persistent state
- `WorkloadSpec` is a runtime type that describes the container setup.
- `RuntimeAdapter` is the main interface with MacOS/Windows/(Linux) implementations that absorbs functionality from exasol-local-vm and actually creates containers and such. This takes a `WorkloadSpec` and manifests it into containers and also checks prerequisites.
    - I think a lot of this is straight from exasol-local-vm except where I moved to `podman kube` and did some cleanup passes.
- `internal/runtimeadapter/types.go`, the `RuntimeAdapter` implementations and the specs are probably good starting points if anyone wants to look at this. Fair warning, I think the specs got away from me a bit while iterating and steering it away from too much backwards compatibility.
    - Looking at this again, I think it I'll introduce a generic podman adapter and have wrappers around that to do the per-platform prerequisites and for MacOs just run it over ssh instead of unattended from a hook. Should be more consistent.

## Type of Change

- [x] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- Move podman container ownership to exasol-personal
- Change podman setup to use declarative `podman kube play`
- Move MacOS database `/exa` data to shared host mount
    - Noticed this when refining the spec. I think this really needs to live on the host because we need to be able to re-create the vm on upgrades.
- Adds amd64 SLC images to manifest, which is all that should be needed for Windows support.
- Probably more

## Testing

### Test Details

**TODO**
- [ ] Need to review unit test coverage.
- [ ] e2e tests should be unaffected
- [ ] Need to add e2e upgrade test for MacOS db data migration

## Checklist

- [ ] Added/updated tests
- [ ] All tests pass locally


moben:
Closing in favor of #248

PR #246, opened 2026-07-30
feat: docker container artifact source SPOT-32024

moben:
## Description

Extends the asset manager to allow embedding container images.

## Type of Change

- [ ] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Add a new download source for `docker://` and `oci:` / `oci-archive:`, the latter two being useful for local testing with any image you want.
- Depends on `skopeo` cli for downloads at build time (related to `podman`, uses the same libraries).
    - Obviously this is only a good idea at build time, but if `embed: true` the resource manager never tries to fetch the resource at runtime so we should be good there
    - We could import it instead but that seemed slightly more annoying because you get to write and maintain part of this: https://github.com/podman-container-tools/skopeo/blob/main/cmd/skopeo/copy.go, but I'm not completely opposed if we think it's better.
- I had to add a deterministic tar archive builder to make checksums stable, based on the go stdlib implementation.
    - An alternative would be to add another interface that "sources" can implement and which lets them supply their own verification and a unique identifier in place of the plain sha256. Thoughts?

## Testing

    - [ ] partial, need to add actual download / local copy tests

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally


moben:
Probably need to add a way to preserve the original url or add some metadata or something so we can know the docker `name:tag` in #248 without needing to parse `podman load` output. We can't make it the filename because that's not safe on windows due to colons.

internal/runtimeartifacts/docker_source.go
```diff
@@ -0,0 +1,145 @@
...
+		}
+	} else if strings.HasPrefix(url, "oci:") || strings.HasPrefix(url, "oci-archive:") {
+		re := regexp.MustCompile(`^(oci:|oci-archive:)([^:@]+)(|:[^@]+)$`)
+		matches := re.FindStringSubmatch(url)
+		if matches != nil {
+			// digest isn't supported for oci/oci-archive
+			scheme, name, tag := matches[1], matches[2], matches[3]
+			return scheme + name + tag, name + tag, nil
+		}
+	}
```
wizenink: for absolute path inputs like `oci:/tmp/foo.tar:latest` this compiles target to `/tmp/foo.tar:latest`.
Later, the crafted command looks like:
`skopeo copy oci:/tmp/foo.tar:latest oci:/tmp/whatever:/tmp/foo.tar:latest`, where skopeo interprets the destiation as:
OCI directory: `/tmp/whatever`
image name: `/foo.tar:latest`
Which is invalid because it begins with `/`.

internal/runtimeartifacts/docker_source.go
```diff
@@ -0,0 +1,145 @@
...
+
+	return "", nil
+}
+
+// Returns the url to copy from using skopeo and the target tag to encode in the
+// oci archive.
+// nolint: revive
+func DockerUrlDownloadTarget(url string) (string, string, error) {
+	if strings.HasPrefix(url, "docker://") {
+		re := regexp.MustCompile(`^(docker://)([^:@]+)(|:[^@]+)(|@sha256:[a-zA-Z0-9]{64})$`)
```
wizenink: This regex drops the port on digest-pinned references. The parser stops name on the first colon, so `docker://registry.domain.com:8080/group/image:tag@sha256:<digest>`, the docker regex produces:
scheme: `docker://`
name: `registry.domain.com`
tag: `:8080/group/image:tag`
digest: `@sha256<digest>`
Because `name` stops at the first colon, the registry port and repo path are treated as part of the tag
We could maybe use https://pkg.go.dev/github.com/novln/docker-parser

internal/runtimeartifacts/docker_source.go
```diff
@@ -0,0 +1,178 @@
...
+	"go.podman.io/image/v5/docker"
+	"go.podman.io/image/v5/docker/reference"
+	ociarchive "go.podman.io/image/v5/oci/archive"
+	ocilayout "go.podman.io/image/v5/oci/layout"
+	"go.podman.io/image/v5/signature"
+	"go.podman.io/image/v5/transports"
+	"go.podman.io/image/v5/types"
+)
+
+type DockerSource struct{}
```
migroble, line 26: I think this should implement the `Identifier` interface to resolve tags to their underlying ID similar to what we do with the git source.

This is done so that if a tag is updated, we consider it a different artifact e.g. if someone wants to use `:latest`, it will get properly updated if there is a new version because the underlying ID changed.

moben: That confused me and I wanted to ask you if that's a bug tbh. I didn't really see `runtimeartifacts` as a feature to always get the latest version of something and it's a bit weird that it can work that way for some sources but not for others.

I can see how it could be useful in some cases though.

internal/runtimeartifacts/docker_source.go
moben: I changed it to use podman's go libraries.

internal/runtimeartifacts/docker_source.go
migroble: I think git and Docker are the only ones whose version identifiers can have aliases. The file source also uses it, but just as a workaround to avoid copying those files into the cache. The HTTP source figures out if something is new by checking if the content hash changed -- I don't quite remember the behavior when there is no content hash (unverified downloads, currently used for presets).

Thinking about it, it might not make sense for Docker as it is right now since it uses a content hash, so fetching a different version without changing that would simply fail.

moben: Now I remember. I originally implemented `Identifier` but that doesn't work with the existing interface because it will check the return value of `Identify` against the actual sha256 of the file. But for container images I can't get the sha256 without downloading the file. I could of course download the image in `Identify` but that seemed to be against the spirit of the interface.

As far as I can tell this accidentally "works" for `GitSource` because directories are skipped when verifying checksums.

We could probably need another interface for "this source does its own consistency check instead of plain sha256", which git could also implement. Though I don't know if there's an API to do that in container-libs and don't particularly want to hand-roll it...

migroble: Yeah, that seems like an issue with the design of the `Identifier` interface, I definitely need to look at this closer.

internal/runtimeartifacts/docker_source.go
```diff
@@ -0,0 +1,191 @@
...
+		return "", fmt.Errorf(
+			"cannot determine destination image for transport %q",
+			source.Transport().Name(),
+		)
+	}
+	if err != nil {
+		return "", fmt.Errorf("load source image descriptor: %w", err)
+	}
+
+	return descriptor.Annotations[imgspecv1.AnnotationRefName], nil
```
loaj84: archiveImageName() reads org.opencontainers.image.ref.name from the OCI descriptor and returns it directly. AFAIR that annotation is optional in OCI layouts, so this can return an empty image name.

Can we fail here with a clear error if the annotation is missing/empty? Otherwise parseImageReference() succeeds and the failure happens later during archive creation/copy.

migroble (dismissed):

loaj84 (changes requested):

internal/runtimeartifacts/docker_source.go
moben: fixed

moben:
Had to rebase to fix conflicts, last two commits are new fixes

loaj84 (approved):

PR #247, opened 2026-07-30
refactor: Add localruntime interface SPOT-32025

moben:
## Description

Adds an interface for the local runtime and moves some things into it to prepare for further changes.

The plan here is that the new interface should be sufficient to allow changing the actual runtime without changes to the surrounding code. Not sure yet if this is sufficient, things like state handling and reporting of ports might need a bit more work.

Possibly needs some more fixes to tests.

## Related Issue

Fixes #

## Type of Change

- [ ] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

Basically see commit list. I'll add motivation here

- New `Runtime` and `VMRuntime` interfaces, the latter is a superset of the former. I'm not yet entirely convinced with the few `switch` on the interface I've added but need to do more work to disentangle health/state/ports and see what will move into `Runtime` and what is actually VM-only
- Moved existing launcher runtime to `localruntime.MacVMRuntime`
- Moved deployment-agnostic version check parts out of `deploy` so it's usable from `localruntime`. Needed to move launcher start parameter construction into `localruntime.MacVMRuntime`

## Testing

### Test Details

Only tested on linux so far but this is just rearranging code so I hope it will just work when tested on mac.

## Checklist

- [ ] Added/updated tests
- [ ] All tests pass locally


internal/deploy/local_reachability.go
```diff
@@ -77,15 +77,20 @@ func classifyLocalReachability(ctx context.Context, runtime *localruntime.Runtim
...
+	case localruntime.VMRuntime:
+		reachabilityErr := classifyLocalReachability(ctx, rt)
+		if reachabilityErr != nil {
+			return reachabilityErr
+		}
 
-	return err
+		return err
+	default:
+		return nil
```
wizenink: Returning nil here makes every failed start, shell or connection operation report success for non-VM runtimes
```suggestion
		return err
```

internal/deploy/local_runtime.go
```diff
@@ -37,68 +37,27 @@ func startLocalRuntime(
...
+		if err != nil {
+			return err
+		}
 
-	clusterIdentity := strings.TrimSpace(launcherState.ClusterIdentity)
-	if clusterIdentity == "" {
-		return nil, errors.New("deployment state is missing cluster identity")
+		return writeLocalRuntimeArtifactsAndWait(ctx, rt, state, waitTimeoutSeconds)
+	default:
+		return nil
```
wizenink: for a runtime implementing only `Runtime`, `Start/0` can return nil, and the deployment is considered successful even if no connection info or secrets exist

wizenink (changes requested):

internal/deploy/local_reachability.go
moben: fixed

internal/deploy/local_runtime.go
moben: Got rid of VMRuntime and should now always `writeLocalRuntimeArtifacts`.

We can probably get rid of a lot of the `VM*` stuff anyway once mac is ported to this.

internal/deploy/local_runtime.go
```diff
@@ -37,41 +37,17 @@ func startLocalRuntime(
...
-	if localConfig.Ports != "" {
-		startArgs = append(startArgs, "--ports", localConfig.Ports)
-	}
-	startArgs = append(startArgs,
-		strconv.Itoa(localConfig.CPUCount),
-		strconv.Itoa(localConfig.MemoryMB),
-		strconv.Itoa(localConfig.DataSizeGB),
-	)
-	if err := runtime.RunCommand(ctx, startArgs, out, outErr); err != nil {
+	if err := runtime.Start(ctx, out, outErr, localConfig); err != nil {
```
moben, line 46: There is a small behavior change here in that errors from the following would also lead to `diagnoseLocalFailure`:
- `os.Remove(paths.StatePath)`
- `localRunnerVersionCheckArgs`
- `localRunnerSlcArgs`

This is technically incorrect but should go away again when the mac launcher is changed to use the podman stuff in #248.

I'd argue this is fine for now because those steps won't really fail in practice but could add a workaround (separate error type probably?)

wizenink: IMO this is fine for now, the release will get #248 merged anyways for the next release and :)

wizenink (approved):

PR #255, opened 2026-08-07
docs: bring README up to date with 2.2 features

juergen-albertsen-exasol:
## Description

The README had not caught up with 2.2. Three features shipped without documentation — named deployments, the `exasol slc` command group, and `exasol diag local` — and the Limitations section still told users that UDFs are unavailable on local deployments, which is no longer true. It was also unclear which of the newer capabilities apply to local deployments, to cloud deployments, or to both.

This PR documents the missing features and makes the local-versus-cloud split explicit wherever it matters.

## Related Issue

None — documentation follow-up to the 2.2 release.

## Type of Change

## Changes Made

- **New section "Deployments and Named Deployments".** Named deployments were only described in a paragraph buried at the end of "Deploy to the Cloud", even though they apply to local deployments too. The section now states up front that it covers both, gives `-d` examples for each (`exasol install local -d demo`, `exasol install aws -d staging`), explains that named deployments live under `~/.exasol/personal/deployments/<name>`, and makes clear that path-based `--deployment-dir` is still supported.
- **Documented `exasol deployments list`** with real output, and corrected a pre-existing inaccuracy: the old text claimed the listing shows "which one is currently active", which it does not. It shows name, status, presets, and path, and covers launcher-managed directories only.
- **New section "UDFs and Script Language Containers".** Leads with the local/cloud difference — cloud deployments come with the standard SLCs; local deployments ship without any — then covers `exasol slc list/install/update/remove`, that the argument is a case-insensitively matched language alias, the database restart with `--auto-approve` / `--no-restart`, and that `install`/`update`/`remove` are local-only.
- **Fixed the Limitations section.** The "UDFs — not available yet (coming soon)" bullet was wrong as of 2.2. It now says UDFs are supported locally once a container is installed, and work out of the box on cloud. Virtual schemas, Admin UI, and single-node are unchanged.
- **Documented `exasol diag local`** as the troubleshooting entry point in the local paragraph of "Start and stop".
- **Cross-references** from Quick Start to the two new sections, so a local-only reader finds them, plus a `-d` hint in "Load Sample Data".
- **CHANGELOG entry** under `Unreleased` → `Changed`.

### Example of the local UDF flow now documented

```bash
exasol slc list                 # available containers and which are installed
exasol slc install python3      # enables PYTHON3 / PYTHON312 UDFs
```

### Example of the named-deployment flow now documented

```bash
exasol install local -d demo    # a local deployment named "demo"
exasol status -d demo           # address it by name from anywhere
exasol deployments list         # see every named deployment and its status
```

## Testing

- [x] Ran `task fmt` and `task lint` — clean (0 issues)
- [ ] Added new tests for new functionality — not applicable, documentation only
- [x] Manually verified every documented claim against the source

### Test Details

Documentation-only change; no code paths are touched. Each behavioral claim was checked against the implementation rather than the changelog:

- `--deployment` / `-d` and `--deployment-dir` are registered together and marked mutually exclusive (`cmd/exasol/commonFlags.go`), and both are present on `install`, `status`, and `connect`.
- `exasol deployments list` output format and the fact that it does not mark an active deployment (`cmd/exasol/deployments.go`).
- SLC aliases (`python3`, `java`, `r`) resolve case-insensitively against the catalog (`assets/resources/slc-catalog.yaml`, `internal/slc/catalog.go`).
- `slc install`/`update`/`remove` are local-only, and the restart-confirmation / `--auto-approve` / `--no-restart` behavior (`internal/deploy/slc.go`, `cmd/exasol/slc.go`).
- `exasol diag local` scope and that it is safe to run when the deployment is down (`cmd/exasol/diagLocal.go`).

Note: `task fmt` and `task lint` were run, but not the full `task all`.

## Checklist

- [x] Code follows the project's coding guidelines
- [ ] Added/updated tests — not applicable
- [x] All tests pass locally — `task fmt` / `task lint` only; no test-affecting changes
- [x] Commit messages follow Conventional Commits format

## Additional Notes

Two deliberate judgment calls worth a reviewer's opinion:

1. **Key Features was left untouched.** Local UDF support is arguably headline-worthy, but the list is already nine bullets, so the local/cloud parity story stays in Limitations instead.
2. **Two pre-existing paragraphs moved sections.** Introducing the "Deployments" heading pulled the paragraphs on preset pinning and the runtime artifact cache out of "Deploy to the Cloud" and into it. Both are generic rather than cloud-specific, so this reads as an improvement — but it is a side effect of the restructure, not an intentional relocation. Easy to move back.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -8,6 +8,8 @@ Notable user-facing changes to Exasol Personal are documented here.
 
 ### Changed
 
+- Documented named deployments, the `exasol slc` command group, and `exasol diag local` in the README, and made clear which features apply to local versus cloud deployments. Named deployments now have their own README section (they apply to both deployment types, not just cloud), a new section covers UDFs and script language containers, and the Limitations section no longer states that UDFs are unavailable on local deployments.
+
 ### Fixed
 
 ### Breaking Changes
--- a/README.md
+++ b/README.md
@@ -74,6 +74,8 @@ exasol install local
 
 In a few seconds you'll have a local Exasol instance running. Run `exasol info` at any time to see how to connect, then head to the **Load Sample Data** section to start querying.
 
+To run UDFs locally, install a script language container — see **UDFs and Script Language Containers** below. To run several local databases side by side, see **Deployments and Named Deployments**.
+
 Prefer to run in your own cloud? See the **Deploy to the Cloud** section below.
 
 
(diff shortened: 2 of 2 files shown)


harish-exasol-india (approved):
LGTM

PR #295, opened 2026-08-21
docs: Clarify contribution guidelines

migroble:
## Description

Clarify repository guidance so contributor-wide practices live in `CONTRIBUTING.md` and LLM-specific instructions live in `AGENTS.md`.

## Type of Change

## Changes Made

- Add code, writing, commit, and OpenSpec contribution guidelines.
- Add agent-specific Git safety guidance.
- Require agents to read the relevant contribution guidelines before planning or editing.

## Checklist

- [x] Documentation updated
- [x] Commit follows Conventional Commits

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -10,7 +10,7 @@ Always read everything that is linked here at least once.
 
 ## Repository Workflow
 
-Before creating a branch, committing, or opening a pull request, read [CONTRIBUTING.md](CONTRIBUTING.md) and follow the contribution process documented there.
+At the start of any task that may modify the repository, read the [Code and Writing Guidelines](CONTRIBUTING.md#code-and-writing-guidelines) before planning or editing. If the task may produce commits, also read the [Commit Guidelines](CONTRIBUTING.md#commit-guidelines) before planning task or commit boundaries. When working with OpenSpec, read the [OpenSpec Guidelines](CONTRIBUTING.md#openspec-guidelines). Follow the rest of the contribution process in [CONTRIBUTING.md](CONTRIBUTING.md) before creating a branch, committing, or opening a pull request.
 
 When preparing a branch or pull request, ask the user for a related Jira or GitHub issue ID if they have not already provided one; when provided, include it in the branch name and pull request description.
 
--- a/CONTRIBUTING.md
+++ b/CONTRIBUTING.md
@@ -11,21 +11,30 @@ Thank you for your interest in contributing! We welcome contributions from the c
 5. **Test your changes** - see [Development Guide](doc/development.md#testing)
 6. **Submit a pull request**
 
+## Code and Writing Guidelines
+
+- **Keep code comments rare and concise** - add comments only for non-obvious current invariants, constraints, or workarounds. Do not explain obvious behavior, repeat names, describe historical changes or rejected alternatives, or refer to tools, tickets, tasks, or OpenSpec artifacts.
+- **Keep task, issue, and pull request descriptions concise and outcome-focused** - put implementation mechanisms in supporting details or acceptance criteria. Avoid unrelated scope, empty sections, and em dashes.
+
+## OpenSpec Guidelines
+
+- **Use positive requirements** - always specify what should happen, never what should not.
+- **Only specify user-visible side-effects** - never include implementation details in requirements. User-visible side-effects include CLI behavior or output, file modifications, network access, process management, etc.
+
+## Commit Guidelines
+
+- **Use [Conventional Commits](https://www.conventionalcommits.org/)** - capitalize the subject after the type and omit the commit body unless the contribution process or a maintainer specifically requests one. For example, `docs: Clarify contribution guidelines`.
+- **Keep cleanup with the related change** - amend the relevant commit instead of adding fixup commits for linting, formatting, typos, or later work that belongs to an earlier task.
+- **Keep OpenSpec tasks independently verifiable** - map each task to exactly one commit and ensure that commit passes the full relevant check suite.
+- **Archive OpenSpec changes before committing their artifacts** - do not commit OpenSpec artifacts while the change is active, and do not mix them with non-OpenSpec files in the same commit. Follow any more specific timing instructions defined by the change.
+
 ## Pull Request Guidelines
 
 - **Keep PRs focused** - one feature or fix per PR
 - **Write clear descriptions** - explain what and why, not just how
-- **Fold cleanup into the related commit** - do not create separate commits for lint, formatting, or typo fixes
 - **Explain CLI changes with examples** - when changing CLI commands or behavior, describe the user-visible behavior in the PR description and include example invocations or output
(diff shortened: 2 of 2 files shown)


AGENTS.md
```diff
@@ -10,12 +10,18 @@ Always read everything that is linked here at least once.
...
 When preparing a branch or pull request, ask the user for a related Jira or GitHub issue ID if they have not already provided one; when provided, include it in the branch name and pull request description.
 
 For non-trivial changes, propose using OpenSpec even if the user has not asked for it; for OpenSpec-backed work, propose archiving the completed change before committing and opening a pull request.
 
+## Git Safety
+
+When using Git to protect work in progress, keep stashes, temporary tags, and temporary branches narrowly scoped. Do not pass `-u` or `-a` to `git stash` unless untracked or ignored files are explicitly in scope.
+
+Inspect the complete contents of a stash, temporary tag, or temporary branch before deleting it. Prefer an isolated worktree for history rewrites when practical.
+
```
moben: I still feel like this isn't repo specific at all and is more of a personal setup.

Maybe something to move to an internal skill?

CONTRIBUTING.md
```diff
@@ -11,21 +11,25 @@ Thank you for your interest in contributing! We welcome contributions from the c
 5. **Test your changes** - see [Development Guide](doc/development.md#testing)
 6. **Submit a pull request**
 
+## Code and Writing Guidelines
+
+- **Keep code comments rare and concise** - add comments only for non-obvious current invariants, constraints, or workarounds. Do not explain obvious behavior, repeat names, describe historical changes or rejected alternatives, or refer to tools, tickets, tasks, or OpenSpec artifacts.
```
moben, line 16: I feel sometimes it's good to have a comment that explains something and also refers to a ticket for extra context.

But keeping in mind that 99.9% of the time this will be read by an LLM and with how much they suck at "sometimes do X" I think I agree with banning it here for the general case.

CONTRIBUTING.md
```diff
@@ -11,21 +11,25 @@ Thank you for your interest in contributing! We welcome contributions from the c
...
+- **Keep task, issue, and pull request descriptions concise and outcome-focused** - put implementation mechanisms in supporting details or acceptance criteria. Avoid unrelated scope, empty sections, and em dashes.
+
+## Commit Guidelines
+
+- **Use [Conventional Commits](https://www.conventionalcommits.org/)** - capitalize the subject after the type and omit the commit body unless the contribution process or a maintainer specifically requests one. For example, `docs: Clarify contribution guidelines`.
+- **Keep cleanup with the related change** - amend the relevant commit instead of adding fixup commits for linting, formatting, typos, or later work that belongs to an earlier task.
+- **Keep OpenSpec tasks independently verifiable** - map each task to exactly one commit and ensure that commit passes the full relevant check suite.
```
moben, line 28: > map each task to exactly one commit

I asked an LLM what it thinks a "task" is and apparently each checkbox in `tasks.md` is a task.
I usually get way more fine grained checkboxes than I want commits (e.g. also something like "run the new test 30× in a loop to make sure it's stable").

moben: Maybe something like these to get at the spirit of the thing?

> - structure changes into self-contained, individually reviewable commits
> - if possible, break large changes down into smaller self-contained changes and sets of commits that build on each other and can be reviewed individually

Don't quite have a good word for changeset/series/... to put in the second one. PR is wrong because this applies when you structure your work and branch is wrong because each branch is the whole history and it's fine to have multiple series on a branch and PR them individually as a stack...

AGENTS.md
migroble: LLMs tend to mess up a lot when rewriting the git history. Since we are asking them to do that a lot, I think it's probably a good idea to tell them how to do it safely too.

CONTRIBUTING.md
migroble: I actually don't like referencing tickets in code, if we ever switch off Jira, all those references are left dangling and future maintainers will be very annoyed at us. This has actually happened to me multiple times and it's never been fun.

Additionally, this is an open source repo, we should avoid referencing things external contributors don't have access to.

CONTRIBUTING.md
migroble: Yes, each checkbox is a task. It might seem overkill, but it makes it trivially verifiable for LLMs. Since each commit will be a lot smaller, we will obviously have a lot more commits, but I don't think that's necessarily a bad thing.

LLMs have incredibly bad judgement: Asking them break large changes down or to structure things in self-contained, reviewable commits is useless in my experience. They have no way of checking whether something is "large" or "reviewable".

I originally tried asking them to commit each "top-level" task and the commits it created were massive and painful to review. When I tried this other approach, it actually structured its task list more closely to what git commits should be. That mapping actually goes both ways.

AGENTS.md
nh-exa: I've not seen that problem that much and indeed don't think it belongs in the repository guidelines because it is not rule for this repository how anybody uses their tools.
You can put it in `~/.claude/CLAUDE.md` and `/.codex/AGENTS.md` and it applies to all work with agents.

CONTRIBUTING.md
nh-exa: I'm also against issue references. I'd rather condense info into a comment if that is required. That would be the "why" of implementation (why does it exist and is the way it is).

```
Keep comments rare, concise, and current. Use them only to explain why non-obvious invariants, constraints, or workarounds exist. Don’t restate the code, document history or rejected alternatives, or reference tools, tickets, tasks, or OpenSpec artifacts.
```

CONTRIBUTING.md
nh-exa: I've checked tasks in a few old specs and we' have 15 commits on average for each user story.

In theory, I don't mind a larger number of commits but I see a conflict in the combination with the requirement that each commit must be standalone. I think for a good number of tasks, that might not the case. I've had an LLM make some analysis of this approach for a few historic specs and it pretty much told me that this would not work for most tasks. They are not "standalone" enough.

You'll force the LLMs to introduce some sort compatibility layers because it will add feature to some part of the system in one task but not update the other part of the system yet because that is next tasks.

It will mean that after each of those 15+ tasks, it will have to run linters and tests. Significantly raising time and token costs.

nh-exa: Therefore, I'm against adding this requirement. Unless the LLM magically comes up with better tasks.

nh-exa (changes requested):

CONTRIBUTING.md
migroble: As mentioned, and to my surprise, the LLM does magically come up with better tasks. I have seen it explicitly mention "write commit-sized tasks" in its chain of thought while creating proposals.

nh-exa: The feature sections, not the tasks might the correct boundary for commits though.

nh-exa (approved):
This is going to be nice improvement

PR #308, opened 2026-08-27
fix: Use Ubuntu 24.04 cloud images for Quadlet-capable Podman

nh-exa:
## Description

Data Lakehouse Turbo activation still fails on cloud deployments after #288. That PR added `podman` to the cloud-init package list, but Ubuntu 22.04 only offers Podman 3.4, which predates the Quadlet systemd generator introduced in Podman 4.4. The package was installed, but the capability the feature actually needs was never present.

This PR moves the provisioned image on all four cloud providers to Ubuntu 24.04 LTS, which offers Podman 4.9 including the Quadlet generator.

## Related Issue

SPOT-32133 (https://exasol.atlassian.net/browse/SPOT-32133)

Reported downstream as L3-4379.

## Type of Change

- [x] `fix`: Bug fix

## Changes Made

- Bumped the provisioned image to Ubuntu 24.04 LTS on all four providers: AWS (`noble`), Azure (`0001-com-ubuntu-server-noble` / `24_04-lts-gen2`), Exoscale (`Linux Ubuntu 24.04 LTS 64-bit`), and STACKIT (`24.04`).
- Corrected the AWS AMI name filter prefix from `hvm-ssd` to `hvm-ssd-gp3`. Canonical publishes releases from 23.10 onwards under the new prefix, so changing the codename alone would have made the AMI lookup match nothing and fail the deployment.
- Updated the top-level README and all four provider READMEs, which documented 22.04.
- Rewrote the CHANGELOG `Fixed` entry from #288, which described a fix that does not work, and added a `Changed` entry for the OS bump.
- Replaced the Exoscale private-network setup, which called `dhclient`. `ubuntu-minimal` depended on `isc-dhcp-client` up to 22.04 but switched to `dhcpcd-base` in 24.04, so `/sbin/dhclient` no longer exists and the unit died with `status=203/EXEC`. The interface is now declared in netplan, which systemd-networkd applies on every boot, and the custom unit is gone.
- Made two pre-existing hang paths fail loudly instead (separate commits, see below).

## Commits

1. `fix: Use Ubuntu 24.04 cloud images for Quadlet-capable Podman`: the image bump itself.
2. `fix: Configure the Exoscale private network without dhclient`: required for 1 to work on Exoscale.
3. `fix: Fail the installation when SSH host key pre-seeding does not succeed`: pre-existing bug, kept standalone.
4. `fix: Bound the node barrier retries during cloud installation`: pre-existing bug, kept standalone.

Commits 3 and 4 are not needed to make 24.04 work, but they are why this took so long to diagnose. `uploadAdminKey.sh` ran `ssh-keyscan ... || true`, so an unreachable node left `known_hosts` empty and every later SSH failed host key verification. `waitNodesReady.sh` then retried that forever with no bound and no connect timeout. The result was a deployment that sat silently for 40 minutes on a one-line error. Both now fail with a diagnostic; they affect every provider, not just Exoscale.

## Testing

- [ ] Manually tested the changes (not possible locally, see below)

### Test Details

`task all` passes (117 passed, 6 skipped). `task fmt` and `task lint` are clean.

No local suite can exercise this change, because it only manifests on a real cloud deployment. The Podman and Quadlet claims were verified against Canonical's published package metadata:

| Release | Podman | Quadlet |
|---|---|---|
| 22.04 Jammy | 3.4.4 | absent, needs 4.4 or newer |
| 24.04 Noble | 4.9.3 | present |

Noble's `podman` package ships `/usr/libexec/podman/quadlet` plus the `podman-system-generator` and `podman-user-generator` symlinks. That last detail matters, because Debian and Ubuntu packaging historically shipped Podman without the generator symlinks, which would have left Quadlet non-functional even on a new enough version.

Provider image identifiers are verified against the live provider APIs:

- **AWS**: the exact Terraform filter resolves in the CI region `eu-west-2` to `ubuntu/images/hvm-ssd-gp3/ubuntu-noble-24.04-amd64-server-20260714` (`ami-01029182857b20417`, x86_64).
- **Azure**: `Canonical:ubuntu-24_04-lts:server:latest` resolves in `westeurope` to 24.04.202608070, `hyperVGeneration: V2`, which is the same generation as the jammy Gen2 SKU it replaces, so there is no VM size implication. Note that Canonical dropped the generation suffix from the offer and SKU naming at 24.04: there is no `24_04-lts-gen2` SKU, and the `0001-com-ubuntu-server-<codename>` offer pattern ends at 22.04. An earlier revision of this branch used the old pattern and the Azure deployment test failed with `PlatformImageNotFound`; the identifiers above are now confirmed against `az vm image list-offers` / `list-skus` / `show`.
- **Exoscale**: diagnosed on a live 24.04 deployment host. `eth1` was `DOWN` with no address, `50-cloud-init.yaml` covered only `eth0`, and `172.16.0.20` resolved via the public default gateway, leaving SSH in `SYN-SENT`. Applying the netplan configuration from this PR brought `eth1` up with the Terraform-assigned `172.16.0.20/22`; after re-running the keyscan the barrier marker appeared in about 10 seconds and the installation proceeded. The same host confirms the point of the PR: `podman version 4.9.3` with `/usr/lib/systemd/system-generators/podman-system-generator -> /usr/libexec/podman/quadlet`.
- **STACKIT**: covered by the deployment tests on this PR.

Only Exoscale needed a change because it is the only provider where the guest configures a second NIC. AWS uses one ENI, Azure one `azurerm_network_interface` with a single `primary` ip_configuration, and STACKIT one `stackit_network_interface`; on those the private address is on the primary interface, which cloud-init configures on any Ubuntu release.

### Live deployment evidence

**AWS deployed successfully on Ubuntu 24.04.** A cloud deployment test provisioned `ami-01029182857b20417`, the Noble AMI the corrected filter resolves to, installed Exasol, and passed `db_connectable`. 26 tests passed in that job.

That job also reported three e2e failures and one error which are not related to this change. They are tracked separately so they do not block review:

- [SPOT-32344](https://exasol.atlassian.net/browse/SPOT-32344): two local Parquet import tests fail because the unquoted table name `Data` is rejected by the SQL parser. The reported error column matches the position of `Data` in all three statements (14, 13, 26).
- [SPOT-32345](https://exasol.atlassian.net/browse/SPOT-32345): `exasol connect` produced no exit hint on stderr.
- The `VpcLimitExceeded` error was an AWS account quota problem and has been resolved.

Both tickets record honestly that no 22.04 baseline exists for this suite, because recent green runs skipped the cloud deployment job entirely. Neither failure has a plausible mechanism tied to the host OS: both are `exasol connect` behaviors against a database that was already up.

Azure and Exoscale still need a clean run. Azure failed on an earlier revision of this branch and is fixed but unverified, and the Exoscale job on that run was manually unblocked on the host during diagnosis, so its result must not be read as a pass.

## Checklist

- [ ] Added/updated tests
- [x] All tests pass locally

## Additional Notes

Deployments created before this change keep running 22.04 and must be recreated to pick up the fix.

Podman 4.9 covers the 4.x Quadlet unit types (`.container`, `.volume`, `.network`, `.kube`) but not `.pod` or `.build`, which arrived in Podman 5.0. If Data Lakehouse Turbo emits those newer unit types, this needs Ubuntu 26.04 (Podman 5.7) instead. That could not be determined from this repository, because the activation logic is product-side and neither SPOT-32133 nor L3-4379 records the underlying Quadlet error.

Unchanged and still out of scope for SPOT-32133: `static.yaml` touches its cloud-init completion marker unconditionally, so a failed package install still reports a healthy deployment. Deployment regression coverage remains deferred to the broader test coverage discussion.

[SPOT-32344]: https://exasol.atlassian.net/browse/SPOT-32344?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


moben (dismissed):

migroble (commented):
Functionally it looks fine, but I'm not a huge fan of the comments and documentation added. 

Comments and documentation should describe the current state of the code, not what changed with respect to a previous version. Additionally, comments should not describe in English what is being "said" in Bash/Terraform/YAML/etc.

PR #315, opened 2026-08-31
fix(runtime-artifacts): Restore embedded preset resolution

juergen-albertsen-exasol:
## Description

Embedded infrastructure and installation presets were completely unavailable: `exasol presets list` reported none, and commands that resolve a preset, such as `exasol install local` and `exasol presets export`, could not find one.

The resource embedder packs a glob group's directory into a tarball, marking directory entries with the trailing separator tar requires. `extractTarEntry` passed that name straight to `os.Root.MkdirAll`, which rejects it, so extraction aborted on the first entry and left an empty directory behind. Glob member lookup then matched nothing, and every caller reported the preset as unknown.

Both halves that combine into this landed together in 1a2249c: presets moved onto the runtime artifact glob mechanism, so a generated tarball reached `TarGzExtractor` for the first time, and that same commit hardened the extractor with `os.OpenRoot`. The previous `os.MkdirAll` tolerated a trailing separator.

Fixing the extractor rather than the embedder is deliberate: trailing separators are the tar convention, so any externally produced archive would hit the same failure.

The second commit makes the cache recover on its own. `getCacheEntry` validated a cached entry with an existence check, which an empty extraction directory satisfies, so a cache written by an affected build kept serving no content indefinitely. Extracted entries now also have to be non-empty to count as a hit.

## Related Issue

No issue ID.

## Type of Change

- [x] `fix`: Bug fix

## Changes Made

- Clear the trailing separator from a tar entry name before it reaches `os.Root`, so directory entries extract
- Treat a cached entry whose extraction directory is empty as a cache miss, limited to entries that declare extraction
- Cover both with regression tests, each verified to fail without its fix

## Testing

### Test Details

Verified against the compiled CLI from `task build` with a cleared cache. On `1a4b2ba` it printed `(none)` for both catalogs; with the fix all five infrastructure presets, both installation presets, and the compatibility matrix render. Beyond listing, which only reads manifests, the extracted cache tree and the output of `exasol presets export aws` are both byte-identical to `assets/infrastructure` under `diff -r`.

Self-healing was verified end to end by emptying the cached `unpack/` directory and rerunning the compiled CLI: presets list correctly and the directory is repopulated. Before the second commit that same sequence stayed broken.

`TestTarGzExtractor_Extract_MultipleFiles` already failed on `main` with `mkdirat sub/: no such file or directory`, so this was a red suite rather than a coverage gap. Across the four affected packages, failures drop from 28 to 9 with no test newly failing. The 9 remaining fail identically before and after: they are pre-existing macOS `/var` versus `/private/var` TMPDIR path comparisons, unrelated to this change.

`task lint-golang` reports 0 issues. `task fmt` fails on `fmt-ruff` for a Python file on `main` as well, untouched here.

## Checklist

- [x] Added/updated tests
- [x] All tests pass locally

## Additional Notes

Anyone who ran an affected build has a cache holding an empty extraction. The second commit means it repairs itself, so no manual cache removal is needed.

`TestDiagnoseLocalUnsafe_VMRunningReportsPortsAndHealth` failed once while checking and passes on repeated runs. It probes real local VM ports, so it is flaky on machine state and unrelated to this change.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -42,6 +42,7 @@ Notable user-facing changes to Exasol Personal are documented here.
 
 ### Fixed
 
+- Fixed infrastructure and installation presets being unavailable. `exasol presets list` reported no presets, and commands that resolve one, such as `exasol install local` and `exasol presets export`, could not find it. Embedded preset archives are now extracted completely.
 - Fixed `exasol connect` omitting its shell exit hint when standard input was piped and emitting it for interactive `--json` sessions. The hint now follows the CLI output contract: it is written to stderr for text shell sessions and suppressed under `--json`.
 - Fixed Data Lakehouse Turbo activation failing on cloud deployments. Podman was missing from the deployment hosts, so activating the feature in the AdminUI hung on a spinner and reported `Activation status for database Exasol could not be retrieved`. Podman is now installed during cloud-init on all supported cloud providers (AWS, Azure, Exoscale, STACKIT).
 - Local runtime host preparation no longer proceeds without approval when a command cannot prompt. Previously a non-interactive invocation was treated as consent, so a scripted run could install Podman unattended. Such runs now fail and explain how to proceed; pass `--auto-approve` for unattended setup.
--- a/internal/runtimeartifacts/manager.go
+++ b/internal/runtimeartifacts/manager.go
@@ -371,7 +371,7 @@ func (m *Manager) resolveUnderLock(
 	}
 
 	if strings.TrimSpace(artifact.Sha256) == "" {
-		slog.Info(
+		slog.Debug(
 			"re-fetching resource without checksum, result may not be stable",
 			"id",
 			resourceID,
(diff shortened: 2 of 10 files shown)


migroble:
Took over this PR. A few updates since the description was written, which I can't edit:

- Dropped the cache self-heal commit (empty extracted dir treated as a miss). Splitting it into its own PR, out of scope here. The "Additional Notes" self-healing claim and its testing note above go with that commit, not this PR anymore.
- Added a small commit moving routine resource-resolution logs from info to debug.
- Dug a bit more into the root cause: it's a narrow Go toolchain regression, reproduces on go1.26.5, not on go1.26.3 (our floor) or go1.27.0, and reproduces the same way on Linux, not macOS-specific. Also found the failure showed up silently because the preset catalog skips a preset whose manifest fails to read without logging it, separately from the extraction issue itself.
- Added an OpenSpec change documenting this fix.

nh-exa (approved):

PR #331, opened 2026-09-03
Assign stable ports to local services

moben:
## Description

Persist concrete host-port mappings for local services so deployments retain predictable endpoints across restarts. Automatic allocation now selects an available port deterministically during initialization or reconfiguration; legacy automatic mappings are migrated before the next safe start.

## Related Issue

[SPOT-32276](https://exasol.atlassian.net/browse/SPOT-32276)

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Select and persist deterministic concrete ports for automatic local-service mappings.
- Bind persisted ports exactly, preserve them across restarts, and provide actionable recovery guidance for bind conflicts.
- Allow stopped local deployments to reconfigure or reset ports and migrate legacy automatic mappings before startup.
- Add portable lifecycle coverage, user documentation, changelog notes, and archived OpenSpec specifications.

## Testing

### Test Details

- Unit and integration suites passed.
- Linux Podman local deployment suite passed all 3 tests in 188 seconds.
- `task all` reaches an unrelated existing Ruff `BLE001` finding at `tests/chaos/test_lifecycle_faults.py:245`.
- The archived OpenSpec change validates successfully. Repository-wide strict OpenSpec validation retains unrelated pre-existing warnings and two incomplete older archives.

## Checklist

- [x] Added/updated tests
- [ ] All tests pass locally

## Additional Notes

Automatic selection occurs when configuration is established or explicitly reset to `auto`, rather than choosing a different port on every normal startup.

[SPOT-32276]: https://exasol.atlassian.net/browse/SPOT-32276?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


nh-exa (dismissed):
I suspect we are going to have to cleanup the boundary of the backend at some point in the near future but this ought to do for now.

nh-exa (dismissed):

nh-exa:
## Deployment test failure: not a flake, and it blocks its own recovery command

Both failing jobs in [run 34098858740](https://github.com/exasol/exasol-personal/actions/runs/34098858740) fail on the same assertion, `tests/tests/deployment/test_local_deployment.py:170` in `_verify_occupied_port_recovery`, after the deliberately failed `start`:

```
INFO root:launcher.py:239 expected status stopped, got status interrupted
```

Linux passed; macOS **and** Windows failed, so this is not a self-hosted-runner timing issue. `interrupted` is a persisted terminal state written by `markOperationInterrupted` (`internal/deploy/deploymentControl.go:157`); nothing later transitions it to `stopped`, so polling for `stopped` would just time out. This run is the only deployment-test run on the branch, so these assertions have never passed on macOS or Windows.

**User-visible impact:** the error tells the user to run `exasol config set --ports ...`, but `interrupted` blocks exactly that command (`deploymentControl.go:80-87`). The recovery path the feature adds is unreachable on two of three platforms.

### Why classification misses, per platform

| Platform | Runtime | Why |
|---|---|---|
| Linux | `HostRuntime` + native podman | Binds the real host port, fails, string match works |
| Windows | `HostRuntime` + podman machine (WSL) | Podman binds *inside* the VM where the port is free, so the container really starts. `podman_install.go:266` returns on `status.Running` **before any string matching** |
| macOS | `MacVMRuntime` (not podman) | Separate `ClassifyBindFailure` call at `mac_vm_runtime.go:218`, additionally guarded by `runnerReportedEndpoint` |

Adding diagnostic markers to `bindConflictDiagnostic` alone would not fix Windows.

### What I would change

1. **Derive the state from a post-failure runtime probe, not from the diagnostic string.** Runtime not running gives `stopped`; genuinely indeterminate gives `interrupted`. *Rationale:* one fragile string match currently decides both state and guidance, so an unrecognized message costs the user a wedged state instead of just a missing hint. A probe is platform-independent and needs no per-runtime markers.
2. **Reduce `ClassifyBindFailure` to guidance only:** whether to append the two `exasol config set --ports` hints. *Rationale:* worst case on a miss becomes a lost hint, never a blocked recovery.
3. **Treat "container running in the VM, host port unreachable" as its own condition** rather than letting the `status.Running` guard bail silently. *Rationale:* on podman machine this is the normal shape of a host-port conflict, not an unclassifiable failure.
4. **Stop using `interrupted` for clean pre-flight failures.** *Rationale:* its message ("Interrupted during start. Please run `start` or `stop`.") is wrong when nothing started, and `markOperationInterrupted`'s own comment concedes it is a catch-all picked to escape `operation_in_progress`, not because it fits.
5. **Log the raw runtime diagnostic at ERROR level.** *Rationale:* the raw text already reaches stderr, but classification misses are currently invisible in CI and telemetry. This run swallowed it via `capture_output=True`, which is why the actual macOS/Windows bind text is still unknown.
6. **Reorder the assertions in `_verify_occupied_port_recovery`** so the diagnostic and actual status are reported before `has_status`. *Rationale:* cheap, masks nothing, and makes the next run hand us the real diagnostic text.
7. **Add `stopped` and `running` to the documented status values** in `statusCmdLongDesc` (`cmd/exasol/status.go:30-36`). Both are missing today.

Items 1 to 4 change the behaviour specified under "Local deployments SHALL bind the persisted service ports exactly", so they likely want an OpenSpec change rather than a drive-by fix.

### Unrelated

The `Can't unlink already-existing object: Permission denied` noise on the macOS job is Go module cache ownership on the self-hosted runner. It occurs during cache restore, before the tests, and Windows failed without it. Worth its own ticket.

nh-exa:
## Correction to the above

Two of the changes I proposed should not be made, and one citation was wrong.

**Retracting items 2 and 3.** I proposed reducing `ClassifyBindFailure` to a host-port bind probe and treating "container running in the VM, host port unbindable" as a conflict. Both are forms of post-failure port probing, which this branch already had and deliberately removed in 96d7afa, along with the `available func() bool` parameter and the `IsAvailable` helper. The rationale is recorded in `openspec/changes/archive/2026-09-04-harden-static-local-port-recovery/design.md:53`:

> A preflight check was rejected because it cannot diagnose a collision occurring after the check. A post-failure probe was rejected because a busy port after failure does not establish why the launch failed.

That reasoning holds, so those items are withdrawn. I had not checked the history before proposing them.

**Item 1 stands, but the wording invited exactly that confusion.** What the state should be derived from is a *runtime liveness* probe, `Runtime.Status`, which asks whether the runtime is running. It is not a port probe and makes no claim about the cause of the failure. The two are unrelated, and only the port probe was rejected.

**The same design note is what makes item 1 necessary.** It accepts the marker-list tradeoff on the explicit premise that an unrecognized diagnostic "falls back to the ordinary launch error without corrupting state". That premise does not hold: the fallback is `interrupted`, and `WorkflowStatePermitsConfigure` refuses `exasol config set` in that state, so the fallback blocks the recovery the failure calls for.

**Citation fix.** I pointed at `deploymentControl.go:80-87` for `config set` being blocked. That is the start-permission code. The actual gate is `WorkflowStatePermitsConfigure` in `internal/deploy/configure.go:52-74`, which permits only `initialized` for any backend and `stopped` for local. The conclusion is unchanged, and it also settles which state a failed local start should leave behind: `stopped` is the only post-deploy state that permits local reconfiguration.

Items 1, 4, 5, 6 and 7 are implemented in https://github.com/exasol/exasol-personal/pull/355, based on this branch.

nh-exa (approved):

PR #332, opened 2026-09-03
feat(local): Store Nano data in host-visible deployment storage

migroble:
## Description

macOS local deployments now keep the Nano `/exa` tree at `local/runtime/exa`, making persistent database files visible to host tools. Linux and Windows keep their existing host-side storage.

Existing deployments migrate on their first start with this launcher. That start also rebuilds the VM guest when it came from a different runner, so the migration and later starts run on the guest the launcher expects.

## Related Issue

[SPOT-32427](https://exasol.atlassian.net/browse/SPOT-32427)

## Type of Change

- [x] `feat`: New feature
- [ ] `fix`: Bug fix

## Changes Made

- Store macOS Nano data at `local/runtime/exa` on the host, migrating existing data on first start and preserving metadata and sparse files.
- Rebuild a deployment's VM guest when the resolved runner differs from the one recorded for it, stopping a running VM first and recording the new runner only once the rebuild succeeds.
- Describe the macOS data-size setting as VM runtime disk size for Podman images and runtime state, since the host filesystem now determines the space available to `/exa`.
- Cover macOS and Windows host-data behavior with tests, and document the host path for virtual-schema setup.

## Testing

### Test Details

Verified on a macOS Apple Silicon host against deployments installed by released launchers:

- A v2.2.0 deployment started on this branch rebuilds its guest, migrates `/exa` to the host, and keeps committed rows. Volume files stay sparse: two 1 GiB files occupy 33 MB and 16 KB on the host.
- A v2.1.0 deployment records no runner version at all and is rebuilt on that basis, also keeping committed rows.
- Restarting a migrated deployment rebuilds nothing and keeps its data.
- A rebuild that fails leaves the recorded runner version untouched, and a later start retries it and succeeds.
- After migration the guest keeps a recoverable backup and no longer has the legacy `/var/lib/exa` path.

`exasol start` reports the one-time rebuild:

```
INFO msg="replacing the local VM guest to match the resolved runner"
```

## Checklist

- [x] Updated documentation
- [x] Updated `CHANGELOG.md`
- [x] Added/updated tests
- [x] All tests pass locally

[SPOT-32427]: https://exasol.atlassian.net/browse/SPOT-32427?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/doc/virtual_schemas.md
+++ b/doc/virtual_schemas.md
@@ -85,16 +85,11 @@ DEPLOY_DIR=$(exasol info --json --deployment demo | jq -r '.deploymentDir')
 DEPLOY_DIR=$(exasol info --json --deployment-dir /path/to/deployment | jq -r '.deploymentDir')
 ```
 
-On Linux, Nano's `/exa` is the persistent Podman data directory on the host:
+Nano's `/exa` files are available at `local/runtime/exa` in the deployment directory:
 
 ```bash
 EXA_DATA="$DEPLOY_DIR/local/runtime/exa"
 ```
-
-On macOS, the database runs inside a managed VM. The deployment's
-`local/runtime/vm-shared` directory is mounted in the VM as `/mnt/host`. Use
-`exasol shell host` to copy staged files into the database directories; do not rely on VM IP
-addresses, SSH keys, or SSH ports.
 
 ## Worked example: PostgreSQL JDBC adapter
 
@@ -138,8 +133,6 @@ directory because the adapter reads it for metadata and the ETL layer uses it fo
 
 ### 2. Copy the files into the database
 
-On Linux:
-
 ```bash
 mkdir -p "$EXA_DATA/bucketfs/bfsdefault/default/vs"
 cp "$VS_DIR/$ADAPTER_JAR" "$VS_DIR/$DRIVER_JAR" \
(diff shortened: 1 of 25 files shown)


nh-exa (dismissed):

assets/infrastructure/local/infrastructure.yaml
```diff
@@ -4,7 +4,7 @@ backend: local
 
 local:
   cpuCount: 2
-  dataSizeGB: 100
+  dataSizeGB: 20
```
moben: I wouldn't reduce it this much, container images need to live there and with sidecars and custom SLC that can add up and it should be sparse anyway.

doc/virtual_schemas.md
```diff
@@ -85,17 +85,13 @@ DEPLOY_DIR=$(exasol info --json --deployment demo | jq -r '.deploymentDir')
 DEPLOY_DIR=$(exasol info --json --deployment-dir /path/to/deployment | jq -r '.deploymentDir')
 ```
 
-On Linux, Nano's `/exa` is the persistent Podman data directory on the host:
+On Linux, macOS, and Windows, Nano's `/exa` files are available at `local/runtime/exa` in the
```
moben: This applies to all platforms now, maybe drop the list?

internal/deploy/local_backend.go
```diff
@@ -36,7 +36,7 @@ const (
 	localInfraMemThresholdMB   = 8192
 	localInfraMemoryNoticeText = "Info: For medium to heavy local workloads, " +
 		"consider increasing VM memory to 8-16 GB."
-	localDefaultDataSizeGB    = 100
+	localDefaultDataSizeGB    = 20
```
moben: see above

nh-exa (dismissed):

moben (dismissed):
The actual migration logic looks okay. Not quite sure why we need the symlink, wouldn't atomic rename+sync be enough?

migroble:
> Not quite sure why we need the symlink

That is actually not part of the migration, we symlink the `exa` directory into `vm-shared` because the VM runner mounts `vm-shared` into the VM and I didn't want to have multiple directories we share between host and guest.

nh-exa (approved):

PR #341, opened 2026-09-04
fix(build): Derive launcher version from release tags

migroble:
## Description

Builds took their version from `git describe --tags --abbrev=0`, which returns the nearest reachable tag of any kind. Since the `docs/*` tags were introduced, that resolves to `docs/0.1`, and the build strips the leading character on the assumption of a `v` prefix, stamping the launcher `ocs/0.1`.

The result is a launcher that panics on `exasol version` and refuses every command with a deployment compatibility check:

```
$ exasol version
panic: semver: Parse(ocs/0.1): No Major.Minor.Patch elements found

$ exasol status --json
Error: invalid deployment version "ocs/0.1": No Major.Minor.Patch elements found
```

Restricting the tag match to `v*` makes builds use the latest release tag again.

## Type of Change

- [x] `fix`: Bug fix

## Changes Made

- Match only `v*` tags when deriving the launcher version in the `build` and `build-windows` tasks

## Testing

### DIFF
--- a/Taskfile.yml
+++ b/Taskfile.yml
@@ -102,7 +102,7 @@ tasks:
       CGO_ENABLED: "0"
     cmds:
       - |
-        version=$(git describe --tags --abbrev=0) # First character is "v".
+        version=$(git describe --tags --abbrev=0 --match 'v*') # First character is "v".
         go build \
           -tags=containers_image_openpgp \
           -o {{.BUILD_DIR_NAME}}/{{.BINARY_NAME}} \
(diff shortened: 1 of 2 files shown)


nh-exa (dismissed):

nh-exa:
The Taskfile change is correct and fixes 29 of the 31 integration test failures, but two remain.

`test_version` and `test_version_json` in `tests/tests/integration/test_cli.py` (lines 80 and 102) recompute the expected version with an unpatched `git describe --tags --abbrev=0` and then strip the first character, so they still expect `ocs/0.1` while the binary is now correctly stamped `2.2.0`:

```
FAILED tests/integration/test_cli.py::test_version - assert 'd' == 'v'
FAILED tests/integration/test_cli.py::test_version_json - assert {'version': '2.2.0'} == {'version': 'ocs/0.1'}
```

Applying the same `--match 'v*'` to both call sites makes the suite green. Using `.removeprefix("v")` instead of `[1:]` also lets the now-vacuous "first character is v" assertion go away.

Verified locally against this branch: 2 failed / 124 passed with the PR as-is, 126 passed with the test call sites fixed.

One note on the checklist: `task all` includes `tests-integration`, so it should surface this. Reproducing it does require the `docs/*` tags to be present locally (`git fetch --tags`).

nh-exa (approved):

PR #366, opened 2026-09-09
feat: preset backend specific cli options

Gregory-Bull-Exasol:
## Description

`--tofu-update-lockfile` was registered as a common deploy flag, so it was offered and accepted for local deployments even though only the OpenTofu backend can act on it. Backend-specific options are now declared and interpreted by the backend that supports them, and the CLI offers only the options of the backend behind the selected deployment.

Behaviour for OpenTofu-backed deployments is unchanged. Local presets and local deployments no longer list the option in their help and reject it with an error naming the preset.

```
$ exasol install aws --help
      --tofu-update-lockfile       Allow OpenTofu to update .terraform.lock.hcl during init

$ exasol install local --help
(no --tofu-update-lockfile)

$ exasol install local --tofu-update-lockfile
Error: unsupported deployment option: infrastructure preset "local" does not support --tofu-update-lockfile

$ exasol deploy --deployment-dir ./local-deployment --tofu-update-lockfile
Error: unsupported deployment option: infrastructure preset "Exasol Local" does not support --tofu-update-lockfile
```

## Related Issue

SPOT-32434

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- Added a backend descriptor registry to `internal/deploy`. Each backend declares, next to its own implementation, how it is constructed, which preset configuration variables it exposes, and which deploy options it accepts. Backend kind resolution, backend construction, and preset variable reading became registry lookups instead of three separate switches, so registering one descriptor is enough to add a backend.
- Replaced `DeployOptions.UpdateDependencyLockfile` with `DeployOptions.BackendOptions`, a name-keyed map. The OpenTofu backend declares `tofu-update-lockfile` and reads it in `Deploy`; the local backend declares no options. The option name and its help text now exist in one place, beside the code that acts on it.
- Deploy operations validate every supplied option against the selected backend's declaration before recording an operation in progress, so an option a backend never declared cannot reach it and be dropped.
- Added CLI pre-registration for backend options, alongside the existing preset variable flag registration. The backend comes from the preset argument for `install` and from the deployment directory for `deploy`. An option only another backend declares is rejected with an error naming the preset, instead of Cobra's `unknown flag`, which reads like a typo.
- Taught the tolerant pre-parser which backend options are boolean, so `exasol install --tofu-update-lockfile aws` still finds its preset argument now that the flag is registered after the preset scan.
- Removed `registerDeployFlags` and `CommonFlags.DeployTofuUpdateLockfile`.

## Testing

### Test Details

`task all` passes: lint, docs, build, unit tests (race detector enabled) and integration tests (132 passed, 6 skipped). The deployment suites (`task tests-deployment`) were not run, since they provision real infrastructure.

New unit tests in `internal/deploy`:

- Option resolution from a preset reference and from a deployment directory, for both backends.
- Rejection of an option belonging to another backend, acceptance of a declared one, and boolean value parsing including an invalid value.
- Registry invariants that a future backend must satisfy: every descriptor is complete, and each option name is declared by exactly one backend.

New unit tests in `cmd/exasol`:

- A tofu-backed preset registers the option as a boolean flag with its description as usage; a local preset registers nothing.
- A local preset and a local deployment reject the option, naming both the option and the preset.
- Help still renders when an unsupported option is present, and an unresolvable preset leaves pre-registration silent so the command reports it.
- An uninitialized deployment directory keeps the known options parseable, so the initialized-deployment gate reports the real problem.
- Collection returns supplied values and omits unsupplied ones.

New offline integration tests in `tests/tests/integration/test_backend.py` export one infrastructure preset per backend from the same source preset, assert the option appears only in the OpenTofu-backed preset's help, and assert the local-backed preset rejects it by name and initializes nothing. The existing unknown-backend test now shares the manifest fixture they introduced.

Manual verification against a built binary covered the help of `install local`, `install aws`, `init aws` and `deploy` for both backends, the option supplied before and after the preset argument, `--flag` and `--flag=value` forms, and `deploy` against an uninitialized directory.
## Checklist

- [ ] Updated documentation (if applicable) - `user-docs/` for user-facing changes, `doc/` for contributor-facing changes
- [x] Added/updated tests
- [x] All tests pass locally

No documentation update was needed: no document in `user-docs/` or `doc/` mentions the option.

## Additional Notes

Three deliberate decisions worth a reviewer's attention:

- `exasol install --help` without a preset argument no longer mentions the option, because no backend can be consulted yet. This matches how the preset variable flags already behave, and preset-specific help is one argument away.
- The error for a `deploy` invocation names the deployment manifest's display name rather than a preset id, which is the only label an initialized deployment carries. `config set` reports preset labels the same way.
- Option declarations are resolved from a manifest alone rather than from a constructed backend. Constructing the local backend resolves a platform runtime and can fail, and help must render on any platform.

The new `cmd/exasol` tests run sequentially, like the other pre-registration tests: recognizing the invoked command goes through the package-global `rootCmd`, and Cobra's lookup mutates shared state that the race detector rightly reports when two tests do it at once.

Option ownership is intentionally scoped to deploy options. The four local-specific behavioural checks in `internal/deploy` are deployment policy rather than option ownership, and were left unchanged. `StartOptions` has no backend-specific options today and can adopt the same mechanism without redesign.

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -57,6 +57,11 @@ Notable user-facing changes to Exasol Personal are documented here.
 
   Example: `exasol install local --help` reports "Infrastructure preset `local`" with `Compatible installation presets: local`, and suggests `exasol install local local`.
 
+- `--tofu-update-lockfile` is now offered only by deployments whose infrastructure preset uses the OpenTofu backend. It stays available for OpenTofu-backed `exasol install` and `exasol deploy`, where help lists it under its own heading, is no longer listed in the help of a local preset or deployment, and is rejected with an error naming the preset when passed to one. On `exasol install` it must follow the preset argument, which is what selects the backend.
+- `--tofu-update-lockfile` is now offered only by deployments whose infrastructure preset uses the OpenTofu backend. It stays available for OpenTofu-backed `exasol install` and `exasol deploy`, is no longer listed in the help of a local preset or deployment, and is rejected with an error naming the preset when passed to one.
+
+  Example: `exasol install local --tofu-update-lockfile` reports `unsupported deployment option: infrastructure preset "local" does not support --tofu-update-lockfile`.
+
 - Local deployments now select and persist a concrete database port during initialization, keep it stable across restarts, allow port changes while stopped, and provide actionable recovery commands when the configured port is unavailable.
 - Documented named deployments, the `exasol slc` command group, and `exasol diag local` in the README, and made clear which features apply to local versus cloud deployments. Named deployments now have their own README section (they apply to both deployment types, not just cloud), a new section covers UDFs and script language containers, and the Limitations section no longer states that UDFs are unavailable on local deployments.
 - macOS local deployments now run the same Podman installation used on Linux inside a managed VM. The VM launcher is responsible only for VM lifecycle, port forwarding, shared files, and command execution; deployment state no longer exposes its SSH transport.
--- a/cmd/exasol/backend_options.go
+++ b/cmd/exasol/backend_options.go
@@ -1,0 +1,252 @@
+// Copyright 2026 Exasol AG
+// SPDX-License-Identifier: MIT
+
+package main
+
+import (
+	"context"
+	"errors"
+	"fmt"
+	"slices"
+	"strings"
+
+	"github.com/exasol/exasol-personal/internal/deploy"
+	"github.com/spf13/cobra"
+)
+
+// Recorded per command, so one command's backend options stay out of another's.
+const (
+	backendOptionFlagsAnnotationKey       = "exasol.backendOptionFlags"
+	backendOptionPresetLabelAnnotationKey = "exasol.backendOptionPresetLabel"
+)
+
+var errBackendOptionFlagConflict = errors.New("backend option flag name conflict")
+
+// Runs before Cobra parses, which is what keeps an option out of the help and the
+// accepted flags of a deployment whose backend cannot act on it.
+func prepareBackendOptionFlags(ctx context.Context, args []string) error {
+	cmd, _ := preregisteredCommand(args)
+
+	switch cmd {
+	case installCmd:
+		return prepareInstallBackendOptionFlags(ctx, installCmd, args)
+	case deployCmd:
+		return prepareDeployBackendOptionFlags(deployCmd, args)
+	default:
+		return nil
+	}
+}
+
+func prepareInstallBackendOptionFlags(
+	ctx context.Context,
+	cmd *cobra.Command,
+	args []string,
+) error {
+	resolution, ok := resolveInstallBackendOptions(ctx, args)
+	if !ok {
+		// The command itself reports an unusable preset, and can list the valid ones.
+		return nil
+	}
+
+	return applyBackendOptionFlags(cmd, args, resolution)
+}
+
+func resolveInstallBackendOptions(
+	ctx context.Context,
+	args []string,
+) (deploy.DeployOptionResolution, bool) {
+	preset, err := scanInfrastructurePresetSelection(ctx, args)
+	if err != nil || preset == nil {
+		return deploy.DeployOptionResolution{}, false
+	}
+
+	resolution, err := deploy.ResolveDeployOptions(ctx, *preset)
+	if err != nil {
+		return deploy.DeployOptionResolution{}, false
+	}
+
+	return resolution, true
+}
+
+func prepareDeployBackendOptionFlags(cmd *cobra.Command, args []string) error {
+	resolution, ok := resolveDeploymentBackendOptions(args)
+	if ok {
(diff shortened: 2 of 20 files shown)


migroble (commented):
Looks alright, but as usual with LLM-generated code, I'm not a huge fan of the amount, size, and content of the comments. I find asking them to do a cleanup pass on comments with the following guidelines works fairly well:

```
- Comments explain **why**, never **what**. The code already says what it does.
- Keep comments rare and concise. Add one only for a non-obvious current
  invariant, constraint, or workaround.
- A comment should nearly never be longer than the code it annotates.
- Remove doc comments on small, self-explanatory functions rather than
  restating the signature. For example, delete "CacheRootPath returns the cache
  root path" above a one-line function called `CacheRootPath`.
- Keep, but tighten, comments that convey something a reader cannot get from
  the code alone: why a legacy-only helper exists, a platform-specific behavior
  split, a non-obvious return-value contract, an atomicity or ordering
  guarantee.
- Delete comments that explain obvious behavior, repeat a name, describe
  historical changes or rejected alternatives, or refer to tools, tickets,
  tasks, or spec artifacts.
- No em dashes. Use a period, comma, colon, or parentheses.
```

nh-exa:
> Taught the tolerant pre-parser which backend options are boolean, so `exasol install --tofu-update-lockfile aws` still finds its preset argument now that the flag is registered after the preset scan.

It sounds like the pre-parser is now more complex because of that. And I think this doesn't need to work and it feels wrong actually to accept an option for a backend before preset is known (`aws` tells which backend should be used).

I think it's fine if its only `exasol install aws --tofu-update-lockfile` that works.

Also, I'd say in help rendering the options with `exasol init --help` it should not appear. But it in `exasol init aws --help` it shall appear in its own group. Reason are the same.

nh-exa:
I think  we should start to view the backend like as if it was the first layer of presets.

If our current preset system and naming wasn't concerned as much about hiding these details from users and a bit more honest, the preset system would actually work something like this.

`exasol init tofu aws c4-asApp` // read as: asApp/c4 Exasol deployment in AWS using tofu
or
`exasol init local podman nano-container` // read as: nano Exasol deployment using podman on local system
or
`exasol init local-macos-vm podman nano-container` // read as: nano Exasol deployment using podman on local MacOs VM 
 
Each layer with its own arguments and capabilities that must be compatible with the next.

nh-exa:
Great rules except:
> - A comment should nearly never be longer than the code it annotates.

I've seen and written plenty of comments that explain why something is done they way its is done that were entire paragraphs long but only referred to a single unusual parameter in e.g. a some Linux syscall.

You can't make rules about comment length like that. @migroble

migroble:
I agree, I've written plenty of novels in comments, but that's why I worded it as "nearly never". LLMs sin of the opposite: Comments are *nearly always* longer than the code they annotate.

Gregory-Bull-Exasol:
> > Taught the tolerant pre-parser which backend options are boolean, so `exasol install --tofu-update-lockfile aws` still finds its preset argument now that the flag is registered after the preset scan.
> 
> It sounds like the pre-parser is now more complex because of that. And I think this doesn't need to work and it feels wrong actually to accept an option for a backend before preset is known (`aws` tells which backend should be used).
> 
> I think it's fine if its only `exasol install aws --tofu-update-lockfile` that works.
> 
> Also, I'd say in help rendering the options with `exasol init --help` it should not appear. But it in `exasol init aws --help` it shall appear in its own group. Reason are the same.

This is in the latest changes. It did simplify it slightly.

CHANGELOG.md
```diff
@@ -85,6 +85,11 @@ Notable user-facing changes to Exasol Personal are documented here.
 
   Example: `exasol install local --help` reports "Infrastructure preset `local`" with `Compatible installation presets: local`, and suggests `exasol install local local`.
 
+- `--tofu-update-lockfile` is now offered only by deployments whose infrastructure preset uses the OpenTofu backend. It stays available for OpenTofu-backed `exasol install` and `exasol deploy`, where help lists it under its own heading, is no longer listed in the help of a local preset or deployment, and is rejected with an error naming the preset when passed to one. On `exasol install` it must follow the preset argument, which is what selects the backend.
```
wizenink:
```suggestion
- `--tofu-update-lockfile` is now offered only by deployments whose infrastructure preset uses the OpenTofu backend. It stays available for OpenTofu-backed `exasol install` and `exasol deploy`, where help lists it under its own heading, is no longer listed in the help of a local preset or deployment, and is rejected with an error naming the preset when passed to one. On `exasol install` it must follow the preset argument, which is what selects the backend.
- 

```

wizenink: this is duplicated

wizenink (changes requested):
Other than the changeling detail, LGTM

PR #376, opened 2026-09-15
fix: allow custom archives without language definitions manifest

loaj84:
## Description

This PR validates available aliases across installed official and custom SLCs and allow template-based custom SLC archives without language_definitions.json.

## Related Issue

[SPOT-32679](https://exasol.atlassian.net/browse/SPOT-32679)

## Type of Change

- [ ] `feat`: New feature
- [x] `fix`: Bug fix

## Changes Made

- Added alias-collision checks between official and custom SLCs.
- Added tests for metadata-less archives and alias conflicts.
- Made language_definitions.json optional for custom SLC archives.
- Updated user documentation and CHANGELOG.md.

## Testing

## Checklist

- [x] Updated documentation (if applicable) - `user-docs/` for user-facing changes, `doc/` for contributor-facing changes
- [x] Added/updated tests
- [x] All tests pass locally

[SPOT-32679]: https://exasol.atlassian.net/browse/SPOT-32679?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


internal/deploy/slc.go
```diff
@@ -687,22 +687,25 @@ func findInstalledByImage(installed []config.InstalledSLC, image string) int {
...
+	providers := make([]slcAliasProvider, 0, len(customs))
 	for _, custom := range customs {
-		for _, official := range officialAliases {
-			if strings.EqualFold(custom.Alias, official) {
-				return fmt.Errorf(
-					"alias %q is currently provided by a custom SLC; remove it with "+
-						"`exasol slc remove %s` before installing this official SLC",
-					strings.ToUpper(official), custom.Alias,
-				)
+		aliases := custom.PackageAliases
```
nh-exa: **Blocking.** This now compares official aliases against `PackageAliases` instead of `custom.Alias`, but `custom.Alias` is the only alias the launcher writes to `SCRIPT_LANGUAGES` (`custom_slc.go:703`). A custom SLC installed under `--alias PYTHON3` from a template archive has empty `PackageAliases`, so `exasol slc install python3` is no longer blocked, and per the comment above this function the alias keeps resolving to the custom container. That is the spec scenario "Official install blocked when a custom SLC owns the alias" (`spec.md:95-99`), and the test that covered it was rewritten to `{Alias: "PY3", PackageAliases: ["PYTHON3"]}`, so nothing catches the loss. Suggest keeping the `custom.Alias` comparison and adding the package-alias providers on top.

internal/deploy/slc.go
```diff
@@ -687,22 +687,25 @@ func findInstalledByImage(installed []config.InstalledSLC, image string) int {
...
+			aliases, err = installedCustomSLCAliases(deployment, custom)
+			if err != nil {
+				return fmt.Errorf("cannot inspect installed custom SLC %q: %w", custom.Alias, err)
 			}
 		}
+		providers = append(providers, slcAliasProvider{Name: custom.Alias, Aliases: aliases})
 	}
 
-	return nil
+	return checkSLCAliasConflicts(officialAliases, providers, "install official", officialName)
```
nh-exa: The spec requires this rejection to guide the user to remove the custom SLC first (`spec.md:71-75`). The previous message named `exasol slc remove <alias>`; the generic one does not, and it identifies the official SLC by flavor, which is not what `slc remove` accepts. The test assertion for the guidance was dropped in the same change.

internal/deploy/slc.go
```diff
@@ -687,22 +687,25 @@ func findInstalledByImage(installed []config.InstalledSLC, image string) int {
...
-			if strings.EqualFold(custom.Alias, official) {
-				return fmt.Errorf(
-					"alias %q is currently provided by a custom SLC; remove it with "+
-						"`exasol slc remove %s` before installing this official SLC",
-					strings.ToUpper(official), custom.Alias,
-				)
+		aliases := custom.PackageAliases
+		if len(aliases) == 0 {
+			var err error
+			aliases, err = installedCustomSLCAliases(deployment, custom)
```
nh-exa: When `PackageAliases` is empty this streams and gunzips the entire custom SLC tarball (`ReadAliases` drains the archive to verify the CRC) just to find zero aliases. Manifest-less archives, which this PR enables, always hit this path, so every `exasol slc install <official>` pays a multi-GB decompress, and it fails outright with `cannot inspect installed custom SLC` if the staged package is gone. Official installs did not touch custom archives before.

Minor: the `len(aliases) == 0` guard above duplicates `installedCustomSLCAliases`, which already returns `PackageAliases` when non-empty.

internal/deploy/custom_slc.go
```diff
@@ -182,6 +220,11 @@ func stageCustomSLCInstall(
 	if err != nil {
 		return nil, nil, fmt.Errorf("invalid custom SLC container: %w", err)
 	}
+	if err := checkSLCAliasConflicts(
+		aliases, officialSLCAliasProviders(state.InstalledSLCs), "install custom", request.alias,
```
nh-exa: This call is what actually rejects the SPOT-32679 repro (official Python installed, custom install of an archive declaring `PYTHON3`), and it has no test. `custom_slc_test.go` covers only the custom-vs-custom and official-vs-custom helpers, and `tests/tests/deployment/test_slc.py` has no alias-conflict case.

user-docs/content/udfs.md
```diff
@@ -70,6 +70,10 @@ A custom package can define one or more database aliases. The launcher rejects a
 those aliases are already provided by another installed custom SLC. Updating an existing custom SLC
 replaces that SLC, so its own aliases may remain unchanged.
 
+Template-based custom SLC archives may omit `language_definitions.json`. In this case, register the
```
nh-exa: The launcher registers `--alias` automatically during activation regardless of the manifest (`custom_slc.go:695-712`), and `spec.md:57-60` requires activation without manual SQL. So manual `ALTER SYSTEM` is only needed for aliases beyond `--alias`, not as stated here or in `CHANGELOG.md:45`.

Also, the paragraph just above still says conflicts arise only against "another installed custom SLC"; this PR extends that to official SLCs.

nh-exa (commented):
Review of `fcca68d`. One blocking finding, the rest are smaller. Full context in the inline comments.

**Specification not updated.** `openspec/specs/local-slc-management/spec.md:101-106` still states "Existing official/custom launcher-alias validation remains unchanged", which this PR changes, and no requirement covers rejecting a custom install whose package aliases collide with an official SLC. Worth an OpenSpec change alongside the code.

**Question:** was replacing the launcher-alias comparison in `checkOfficialAliasNotHeldByCustom` intentional? The agreed approach in SPOT-32679 was collision detection in both directions, which reads as adding the package-alias direction rather than substituting it for the existing one. Non-blocking, but it drives the first inline comment.

loaj84:
> Review of `fcca68d`. One blocking finding, the rest are smaller. Full context in the inline comments.
> 
> **Specification not updated.** `openspec/specs/local-slc-management/spec.md:101-106` still states "Existing official/custom launcher-alias validation remains unchanged", which this PR changes, and no requirement covers rejecting a custom install whose package aliases collide with an official SLC. Worth an OpenSpec change alongside the code.
> 
> **Question:** was replacing the launcher-alias comparison in `checkOfficialAliasNotHeldByCustom` intentional? The agreed approach in SPOT-32679 was collision detection in both directions, which reads as adding the package-alias direction rather than substituting it for the existing one. Non-blocking, but it drives the first inline comment.

All review comments have been addressed.

nh-exa (approved):

PR #377, opened 2026-09-15
docs: Generate the CLI reference from the launcher binary

nh-exa:
## Description

The published user documentation had no command or flag reference, although the product is a CLI. Commands such as `init`, `deploy` and `presets export` were undocumented, as were the preset-specific `install`/`init` flags that control cluster size, volume sizes, passwords and optional object storage.

This adds a CLI reference page generated from the launcher itself, so it cannot go stale.

The generator invokes the built binary and captures its real `--help` output, rather than using `doc.GenMarkdownTree`. That was deliberate: the preset-specific flags are registered at parse time and are invisible to a static tree walk, and this approach needs no changes to the launcher.

## Related Issue

SPOT-32691

## Type of Change

## Changes Made

- `user-docs/scripts/cli_reference.py` walks the command tree via cobra's `__complete`, captures `--help` for all 56 commands, and enumerates presets from `presets list --json`
- `task docs-cli-reference` regenerates the page; `task docs-cli-check` verifies it and is part of `task docs-check`, which `task all` now runs
- `lint-python` gains the Go toolchain, since the check builds the launcher to compare against
- `user-docs/content/cli-reference.md` added to the nav under Help and reference

## Testing

### Test Details

`task docs-check`, `task lint`, `task fmt` and `task tests-unit` all pass locally. `task tests-integration` was not run, as it provisions real deployments; nothing here touches Go code.

Behaviour verified rather than assumed:

- **Drift is detected.** Tampering with a flag description makes `task docs-cli-check` fail with a unified diff and `run 'task docs-cli-reference' and commit the result`; it passes again on restore.
- **Output is reproducible.** Two consecutive runs are byte-identical, and no `/home` or `/tmp` path appears in the page. The binary runs under a temp `HOME` that is substituted for `<home>`.
- **Nothing is dropped.** Every `--flag` token in the raw help of root, all five presets, `status`, `connect`, `slc list` and `cache clean` appears in the page.
- 81 documentation tooling tests pass, covering the help parsing, section filtering, grouping and rendering.

## Checklist

- [ ] Updated `CHANGELOG.md` for user-facing changes
- [x] Added/updated tests
- [x] All tests pass locally

No changelog entry: this changes no product behaviour, and the flags it documents already shipped.

## Additional Notes

**Post-processing.** A verbatim dump ran to 1378 lines, about a third of it repetition. The generator now hoists the global flags into one section (they were repeated across 46 blocks), drops the `--help` navigation trailers, and collapses the cobra shell-completion boilerplate into `<details>`. It also orders commands by the launcher's own groups (Essential / Lifecycle / Additional) instead of alphabetically, and uses `data-toc-label` so the table of contents reads `slc install` rather than `exasol slc install`.

**Fails loudly by design.** If commands ever report differing global flags, or a command falls outside every group, generation stops and names the problem rather than silently publishing something wrong.

**Found while building this, not fixed here:** `exasol install <preset> --help` silently drops the entire preset flag section when `HOME` points at a non-existent directory, printing generic help with no error. That is why the generator uses a writable temp directory. Worth a separate issue.


nh-exa:
Reworked the CI wiring after measuring the first run.

Step timings showed `lint-python` had gone from ~23s to 193s, because it was setting up the Go toolchain (93s) and building the launcher (inside an 82s docs-check) purely to check the CLI reference.

The check now runs in the `build` job, against the launcher that job already produces:

- `lint-python` is Python-only again (no `setup-go`, no `fetch-depth: 0`) and runs `task docs-check-standalone`
- the `build` job's ubuntu leg gains `setup-python` and runs `task docs-cli-verify`

`task docs-check` is unchanged for local use: it is now `docs-check-standalone` plus `docs-cli-check`, so it still builds the launcher and verifies the reference in one command.

Expected effect, based on measured step times: wall-clock gate back to ~160s (from 193s) and compute to ~195s (from 352s).

Verified locally: the coverage-instrumented binary the build job produces generates a byte-identical reference; drift is still caught through both `task docs-cli-verify` and `task docs-check`; `task lint` and `task docs-check` pass.

Taskfile.yml
```diff
@@ -401,8 +426,7 @@ tasks:
     desc: Run all tasks (lint, build, test)
     cmds:
       - task: lint
-      - task: docs-test
-      - task: docs-build
+      - task: docs-check
```
nh-exa: split into docs-cli-check after build so i can use the artifact created by it

wizenink (approved):

